PR #8, opened 2016-07-06
fullwood1 -> logblock

renbear67:
gobal replace of fullwood1 with logblock in all non-dungeon files


sayterdarkwynd:
i like chips

PR #17, opened 2016-07-10
Mining Ores fix

Royal-Jelly:
Adds sounds and particles to Fracking U's Ores ~

sorry about the many small commits ~ I don't have a working client yet

### DIFF
--- a/tiles/mods/algaeore.matmod
+++ b/tiles/mods/algaeore.matmod
@@ -6,7 +6,10 @@
   "health" : 4.5,
   "harvestLevel" : 2,
   "breaksWithTile" : true,
-
+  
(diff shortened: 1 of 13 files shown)


sayterdarkwynd:
i dont really WANT sounds and particles on the FU ores. They are as visible as intended....

Royal-Jelly:
ah i should have figured that was intentional, but  without any sort of indication that you are indeed mining an ore its kinda weird for me

Seeing as you aren't happy with the default ones there is also the option of having ones unique to your ores

sayterdarkwynd:
oooh you meant it was MISSING things. Sorry, I thought it was adding glow or some shit. NEvermind. I checked what you did, and was not aware it was missing :)

PR #25, opened 2016-07-12
Fixed some bugs.

debugman18:
Applied a lazy fix to the portrait crash by giving each and every quest a portraits parameter with a questGiver value.

Converted old gathering quests to the new format introduced in 1.0. 

Fixed ice rain crashing on crystal moons.

I haven't extensively tested these fixes, but I was not able to reproduce crashes afterwards.


sayterdarkwynd:
nice. thanks. Since you're helping with the odd thing here and there, my official checklist is here: https://docs.google.com/spreadsheets/d/1HkPxJucwMXS4Wm_sApm_vgM4XmrRxSCjAbqvtVHFU2Q/edit#gid=2026261600

PR #70, opened 2016-07-28
starbooze tweak - water barrel now takes wheat instead of wheat seeds

draeath:
… seeds as input for soaked wheat (which is then used on the sprouting table). includes new graphics for the sprouted and soaked wheat, to further differentiate them from wheat seeds and wheat itself (eg, to assist those with color blindness). This change is intended to make the wheat input for the chain leading to malt renewable, similar to hops and other input crops. (previously wheat seeds where used. as wheat only drops one seed when harvested, this meant it was a net loss)


draeath:
I have tested this change.

sayterdarkwynd:
nicely done, good sir.

PR #72, opened 2016-07-28
recipe tweak - fixes #69

### LINKED ISSUE #69: Healing I EPP Augment: Uncraftable
The recipe calls for two units of Microchip, which appear to be deprecated.  
Might want to search for instances of that crafting material to clean up any other references to it from cropping up in the future.

draeath:
replace microchip (deprecated) recipe input with advcircuit as current with the other tier1 augments.

### DIFF
--- a/recipes/augments/augment_regen1.recipe
+++ b/recipes/augments/augment_regen1.recipe
@@ -1,6 +1,6 @@
 {
   "input" : [
-    { "item" : "microchip", "count" : 2 },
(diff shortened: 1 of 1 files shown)


renbear67:
Looks good to me. @sayterdarkwynd will have to be the one to merge it, since he's the only one with write access.

sayterdarkwynd:
microchip isn't deprecated, though. Not sure why it would be thought of as deprecated.

PR #76, opened 2016-07-29
fix #65 - adds Frackin' Universe's ores to vanilla ore detector

### LINKED ISSUE #65: Ore Detector not working with FrackinUniverse ores.
Everything is in the title.
To be fair though, it doesn't detect Coal either on an unmodded game, so it's up to you.

draeath:
Does what it says on the tin. Patch in color assignments to the ore detector to support ores added by Frackin' Universe. Colors chosen with a sampled average color of the tile graphics for respective ores.

I couldn't find a way to add a line "second to last" so instead am first removing the last item (none) and placing it back in after the new additions.


renbear67:
By the way, order rarely matters in the JSON files. This might be an exception, though. (This is re: removing/readding "none")

PR #104, opened 2016-08-05
Small Solar Array rebalance

renbear67:
Smoothed out the power curve a bit for the Solar Array. (It was going from 25% -> 100% with just a tiny boost in height.)


renbear67:
<emily>... nevermind... </emily>

PR #112, opened 2016-08-07
I'm an idiot

renbear67:
fat-fingered a period rather than a comma.

I need to take my own advice; THIS IS WHY WE TEST

I could have SWORN I tested


sayterdarkwynd:
already pushed a fix the second i got up lol. 


it happens.

PR #150, opened 2016-08-12
Gardenbot2 ore support

LoPhatKao:
add a patch file to root so gardenbots recognise FU ores
if you've added any ores since 1.0, will need to add those as well 
(or if i fked spelling on any xD)


sayterdarkwynd:
thx dude!

Issue #159, opened 2016-08-15
Machines stay on with leftovers


gamrguy:
If leftovers (less than required amount of a processable material) are left in a machine, the machine will continue to be on and consume power, even though the machine is no longer actually processing anything.

To replicate: Place any amount of processable material into a powered machine that processes in groups that is less than the required amount (e.g. one ore in a blast furnace, 1-2 liquid in a centrifuge). The machine will appear to be on and will consume power while not doing anything.

renbear67:
Fixed in pull request #160

PR #160, opened 2016-08-15
Active consumption + misc

renbear67:
Stations will no longer draw power when there's a "leftover" in their input slot. (1-4 items, too few to process)

Powered centrifuges were erroneously listed in the fourth tab of the MA. This moves them to the "powered stations and components" tab.

Small apiary text and unlock corrections.

vmite description fix. Was referring to an item so old I've never seen it.


renbear67:
Issue #159 is about this problem.

PR #166, opened 2016-08-18
A fix and Biome/block Tweaks

Jeadbe:
==Changes==
--made tidewater floors use its own parallax (also made that parallax
look like a rocky version of the ocean floor parallax)

--changed hot magmatiles to turn into cooled volcano rock on contact
with water
--changed flower crop to produce primarily petals instead of ludicrous
amounts of seeds

==Fixes==
--stoped extraction lab from producing 21 iron ore from rusty metal
--removed unused (?), duplicate tidewater floor images


Jeadbe:
For reference, this is what the tidewater floor would look like now:

sayterdarkwynd:
nice. ill try it out.

PR #181, opened 2016-08-30
Wire relay and battery fixes

NotMrFlibble:
Also make solar panel debug text easier to read.


renbear67:
Looks good!

NotMrFlibble:
There's still an issue if you connect a 5u item (for example) to a pair of 8u batteries – it'll drain 5u per battery tick from each. I've not tried to fix that as yet, but it looks like it'll involve finding out how many distinct power supplies are attached to a given power consumer.

PR #183, opened 2016-08-30
Batteries: improved state display, charge retention while not placed

NotMrFlibble:
Yes, I too have accidentally removed a placed, charged battery.


sayterdarkwynd:
hey dude, question for you:

renbear has lost his net for a month or so and I have some shit that needs fixing. specifically, shield.lua within items/active/  of FrackinRaces. Long story short, trying to get the damageListener to remove the blockCount buff when the player takes damage rather than only if they fail a block. It's easily exploitable for uber bonuses otherwise.

We ran into a rut and coudlnt figure it out. Mind taking a look?

renbear67:
Not just the blockCount buff, the persistentEffects per-race. I cleaned it
up a bit, but the fundamental logic just isn't working. I have a horrible
suspicion that there's some big stupid logic error in the modified
damageListener.

NotMrFlibble:
Yes, logic error. There's a pull request waiting…

PR #187, opened 2016-08-31
Power supply: small optimisations and battery drainage improvement

NotMrFlibble:
Powering a device from more than one battery now properly spreads the load, rather than draining as if it were n devices each connected to a different one of n batteries.

However, for accurate calculation (given the current state of the code), there is a limitation:
- you may connect any number of direct power supplies (batteries and/or generators); or
- you may connect one indirect power supply (wire relay) only..

Power sensors will, when misconnected, indicate ‘invalid’.

`isn_getCurrentPowerInput()` may return `nil`.


sayterdarkwynd:
OOOOH ! awesome!

PR #212, opened 2016-09-07
Terramart Shipments

NotMrFlibble:
Additional food, drink and seeds will all be recognised automatically so long as their categories are correctly set. Additional textiles and monster drops will be recognised by itemTags content (however, none are currently set).


renbear67:
Nice idea. Now, this is an override, right? Do you know if they're changing the vanilla version of this in SB 1.1? It's not a stopper if they are, we'll just have to be careful.

NotMrFlibble:
At present, this file and its corresponding .config file are unchanged in 1.1. Whether it'll remain so, I have no idea.

PR #220, opened 2016-09-09
The Big Bee Patch

Tamamo89:
We're frackin golden guys just like the honey we make. Actually the majority of honey is not golden but that is neither here or there.

*Changelog
Catch and Crossbreed bees to get new Queens to Unlock More Recipes added to the Beestation!
The Industrial Centrifuge and Honey Jarrer now require Golden Glass as well as Golden Planks for the former. They are unlocked by tungstenbars.
The iron bar will unlock the iron centrifuge.
Amber Chunks unlock Mid Tier Weapons.
Godly Combs unlock High Tier Weapons.
Queens unlock the majority of recipes with the exception of the Tier 2-3 Centrifuge and Honey Jarrer which require bars to unlock.
.
*Bug Fixes
Bee Station Filtering Issue has been addressed.
Amber Shield is now uses correct images.

*Suggestion
1: Crystal bees that produce the Honey Crystal Block material for Honey Crystal Furniture.

### DIFF
--- a/interface/windowconfig/beestation.config
+++ b/interface/windowconfig/beestation.config
@@ -1,5 +1,5 @@
 {
-"requiresBlueprint" : false,
+"requiresBlueprint" : true,
   "paneLayout" : {
     "background" : {
       "type" : "background",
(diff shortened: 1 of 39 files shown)


sayterdarkwynd:
nice work. added.

PR #225, opened 2016-09-10
Bugfix

Tamamo89:
Forgot the most basic of the frames. basicframe.item
It is now craftable.


renbear67:
This needs to be merged in ASAP

PR #228, opened 2016-09-11
FU armor text fixes

Misunderstood-Wookiee:
Fixes most of the overlapping text issues when crafting FU armors.


NotMrFlibble:
Cherry-picking 304766c.seems like a good idea here.

renbear67:
Isn't this sort of thing exactly what squash-and-merge is for? That reduces everything down to one commit.

sayterdarkwynd:
yea, lets do that. Less worry that way. Can you do that up, wook?

Misunderstood-Wookiee:
I would squash and merge if I knew how, I only joined Git too help with this mod Renbear was helping me half of yesterday just to understand the basics.

renbear67:
Squash-and-merge is what we do when we click the big green button below. :-) I'll get clearance from @sayterdarkwynd tomorrow.

Misunderstood-Wookiee:
Oh must only be visible in repo that's mine.

Misunderstood-Wookiee:
@NotMrFlibble 
Cherry pick bad idea unless you have changed the file between this pull you need every edit in it or the status won't fit properly.  Sayter also knows he needs to unify the armors to use armornew2 at some point or we can't fix placement for all the armors as some use the vanilla tooltip.

PR #390, opened 2016-10-11
Explosives Updated.

Tamamo89:
Explosives now wirable all that's left to do, is put them on their own circuit, that's a bit beyond my lua knowledge unfortunately. Also need the plunger.


Tamamo89:
Actually I'm going to take a stab at putting them on their own circuit. Shouldn't be too hard.

sayterdarkwynd:
ill wait till its totally done, then.

PR #409, opened 2016-10-22
Mad Scientist

Tamamo89:
Added Mad Scientist Clothes


FullOfSpiders:
Please excuse my impertinence. In the description for the Mad Scientist Chest, genius should be spelled without an o.

Also, I will finally be able to perform mad science without being mistaken for a conventional scientist. Genetic abominations for everyone!

PR #417, opened 2016-10-28
Science Outpost Quest Updated

Tamamo89:
I might of missed a few things however. But hopefully things are as they should be.
tl;dr Do a fetch first and make sure I didn't fuck anything up. If I did, let me know and I'll append.


Tamamo89:
Dummy out old quest don't overwrite them ye damn kitsune!

Tamamo89:
@sayterdarkwynd 
Okay it's been changed over to the way you requested.

PR #418, opened 2016-10-29
Update fu_storageutils.lua

bk3000:
minor efficiency improvement

### DIFF
--- a/scripts/fu_storageutils.lua
+++ b/scripts/fu_storageutils.lua
@@ -6,8 +6,10 @@ function fu_storeItems(items, avoidSlots, spawnLeftovers)
 
 		for i = 0, world.containerSize(entity.id()) do -- iterate all output slots
 			local stack = world.containerItemAt(entity.id(), i) -- get the stack on i
-			if stack and (stack.name == something) then -- not empty and its "something"
-			  table.insert(slots,i) -- possible drop slot
+			if stack then -- not empty
+				if stack.name == something then -- its "something"
+					table.insert(slots,i) -- possible drop slot
+				end
 			else -- empty
 				table.insert(empty, i)
 			end
--- a/scripts/fu_storageutils.lua
+++ b/scripts/fu_storageutils.lua
@@ -1,9 +1,3 @@
---edit notes from bk3000 aka bk3k
---
---I changed a few things mostly from pairs to ipairs for efficiency 
---where there will be no examining anything but iterated lists
---aka nothing like myTable.nonNumber_index
-
 function fu_storeItems(items, avoidSlots, spawnLeftovers)
 	local function fu_getOutputSlotsFor(something)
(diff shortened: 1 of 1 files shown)


NotMrFlibble:
NAK. There's a behaviour change: there are cases when a storage slot should not be added to either list (when it doesn't match what's wanted _and_ it itsn't empty).

And several of the comments belong in the commit message…

scripts/fu_storageutils.lua
```diff
@@ -6,23 +12,21 @@ function fu_storeItems(items, avoidSlots, spawnLeftovers)
 
 		for i = 0, world.containerSize(entity.id()) do -- iterate all output slots
 			local stack = world.containerItemAt(entity.id(), i) -- get the stack on i
-			if stack then -- not empty
-				if stack.name == something then -- its "something"
-					table.insert(slots,i) -- possible drop slot
-				end
+			if stack and (stack.name == something) then -- not empty and its "something"
+			  table.insert(slots,i) -- possible drop slot
```
NotMrFlibble: No. Nested `if` is correct here.

bk3000: if condition1 then if condition2 then end end
will behave identically to 
if condition 1 and condition2 then end
because expressions on the right hand of 'and' will not be checked in the event the left hand is either false or nil.  'or' is the opposite case where in expressions to the right side of 'or' will not be evaluated unless the expression to the left is either 'false' or 'nil'

'If' will evaluate expressions in a boolean fashion and 'nil' is treated as 'not true' just like 'false' where as 'true' and any value that isn't 'false' or 'nil' will be treated as true for the purposes of boolean expression.

It really is equivalent behavior.  But take that part out if it really bothers you.

bk3000:
Maybe I also should explain my line 63.
I passed along this argument to ipairs()
object.getOutputNodeIds(node) or {}

ipairs would receive the return of object.getOutputNodeIds(node) as long as that return is not either 'false' or 'nil'
In the event that object.getOutputNodeIds(node) returns either 'false' or 'nil' then ipairs() will be passed and empty table to avoid errors.  However
do codeblock end
won't be called because there is nothing in the empty table.

essentially that is a quick and easy way to sanitize the return of object.getOutputNodeIds(node)
I'm skipping the temporary memory allocation/assignment/then read of 'out'
It isn't much gain, but savings are savings.

scripts/fu_storageutils.lua
NotMrFlibble: While what you write is true in and of itself, there's an `else` block to consider, and doing the rewriting which you propose changes the condition required for execution of that block from `not condition1` to `not condition1 or not condition2`. The implicit ‘do nothing’ in the case of `stack and stack.name ~= something` _is intended_.

So, for that change, still NAK.

PR #419, opened 2016-10-29
Update isn_solarpanel.lua

bk3000:
efficiency improvement, also request to check that intent of code matches result

### DIFF
--- a/objects/power/isn_solarpanel/isn_solarpanel.lua
+++ b/objects/power/isn_solarpanel/isn_solarpanel.lua
@@ -1,22 +1,12 @@
---edit notes from bk3000 aka bk3k
---
---efficiency improvement by not calculating truepos over and over when the location won't change!
---plus a small improvement in update()
---and make sure line 34 - 36 is as intended... seems like that might not me and could be done more efficiently by
---checking against larger values first combined with if, elseif, elseif, elseif to set the actual desired value
-
 function init(virtual)
 	if virtual == true then return end
 	storage.checkticks = 0
 	storage.truepos = storage.truepos or {entity.position()[1] + math.random(2,3), entity.position()[2] + 1}
-  --if not nil, no reassignment.  If nil then gets assigned
-  --although I honestly do no understand why the random 'x' offset but I left it intact
 end
 
 function update(dt)
 	storage.checkticks = storage.checkticks + 1
-	--if storage.checkticks >= 10 then
-  if storage.checkticks > 9 then --slightly more efficient because >= is equivalent to n > v or n == v
+  if storage.checkticks > 9 then 
 		storage.checkticks = 0
 		isn_getCurrentPowerOutput()
 	end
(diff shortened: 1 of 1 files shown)


objects/power/isn_solarpanel/isn_solarpanel.lua
```diff
@@ -1,48 +1,57 @@
+--edit notes from bk3000 aka bk3k
+--
+--efficiency improvement by not calculating truepos over and over when the location won't change!
+--plus a small improvement in update()
+--and make sure line 34 - 36 is as intended... seems like that might not me and could be done more efficiently by
+--checking against larger values first combined with if, elseif, elseif, elseif to set the actual desired value
+
```
NotMrFlibble: This belongs in the commit message.

bk3000: I haven't used github much.  I'll do so in the future.

bk3000:
`if storage.checkticks > 9 then`
is slightly more efficient because `>=` is equivalent to `n > v or n == v`
>= is two evaluations where as > is one evaluation so i shifted the value it would check against by 1.

also 

```
if light > 0.1 then generated = generated + 0.25 end
if light > 0.3 then generated = generated + 0.35 end
if light > 0.5 then generated = generated + 0.45 end
if light > 0.7 then generated = generated + 0.55 end
```

Just to check your intention here, these 4 checks are not mutually exclusive, so if light > 0.7 then generated is going to actually increase 1.6 in total, not 0.55 thanks to a compound increase.

If that wasn't the end value you where going for, the solar lights are over-performing.  However if you really wanted that result...

```
if light > 0.7 then 
  generated = generated + 1.6
elseif light > 0.5 then
  generated = generated + 1.05
elseif light > 0.3 then
  generated = generated + 0.6
elseif light > 0.1 then
  generated = generated + 0.25
end
```

That will get you there more efficiently.  That code works better anyhow but plug in the desired increases you actually wanted.  However I do not know what increases you where aiming for.

bk3000:
That comment cut out >=

I do not know why.

NotMrFlibble:
`a >= b` is not the same as `a > b - 1`. It happens to work if the two values are guaranteed to be integers, but otherwise offers no improvements other than (where b is an integer literal) reducing script size. (Which you've negated by messing with the indentation.)

The solar panel power output decision I'll leave to Sayter, but I think that I'd prefer something like
`generated = generated + func(light)` rather than multiple `if` statements.

sayterdarkwynd:
so long as the output works, im not overly picky on how it does so

bk3000:
I'm aware that doesn't work quite the same for float numbers.  But your ticks are integers.  Likewise if you where doing a series of a table indexes... that's an integer so the same concept applies.

Of course >= is the best way when you don't know in advance if you're dealing with integers.  And getting used to that way could prevent a human mistake when you forgot that you may deal with floats.  So your view has merit.

But in this case it is known to be an integer and I'm just trying to squeeze a tiny amount of efficiency.  Efficiency is often at odds with versatility... but not in this exact case.

Sorry if I'm being a pain.

sayterdarkwynd:
you arent.

 Sorry for the delay: Still waiting on Flibble to let me know if these are good to merge or not.

PR #420, opened 2016-10-29
Update isn_geldetection.lua

bk3000:
small efficiency improvement


bk3000:
You added this script to the player scripts, so small efficiency improvements probably count for something.

PR #421, opened 2016-10-29
Update

bk3000:
Efficiency improvement by avoiding needless recalculating every update() tick

### DIFF
--- a/objects/fu_watcher/fu_watcher.lua
+++ b/objects/fu_watcher/fu_watcher.lua
@@ -1,10 +1,3 @@
---update notes from bk3000 aka bk3k
---
---objects aren't mobile so why recalculate every tic of update?  Just a waste of CPU cycles doing that.
---I did it in init()  It probably only needs to recalculate between session changes
---since only storage.state persists between sessions.  That being the case, 
---my edit of line 10 will keep the setting between sessions (which I'd find more desirable by far).
-
 function init()
 	object.setInteractive(true)
 	storage.state = storage.state or false
--- a/objects/fu_watcher/fu_watcher.lua
+++ b/objects/fu_watcher/fu_watcher.lua
@@ -1,14 +1,22 @@
+--update notes from bk3000 aka bk3k
+--
+--objects aren't mobile so why recalculate every tic of update?  Just a waste of CPU cycles doing that.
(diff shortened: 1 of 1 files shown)


bk3000:
I also find it VERY strange that SB's rectangles are defined as 
{number, number, number, number}
instead of 
{ {number, number}, {number, number} }
so very counter-intuative like so much else.

sayterdarkwynd:
and this changes....what?

NotMrFlibble:
Indeed. The first line of the commit message needs to be a brief description of the changes; and I think that most, if not all, of the added comments would be better placed in the commit message.

sayterdarkwynd:
are these all good to go, or what? I dont want to commit any lua changes you or renbear do not approve

bk3000:
Sure thing.  I'm just accustomed to throwing the comments in but no reason to not cull them if you don't need them.  I haven't really used github much either so forgive me on that.

I'm not objected to commenting in this space instead.

bk3000:
`watchBox` is not calculated anymore each `update()` tick because the location of objects is static.
`storage.watchBox` is calculated in `init()` instead and referenced where `watchBox `was once referenced.

In the event that `storage.watchBox` has already been defined, it will not be calculated again.  For example when you beam up and down to/from your ship.  the `storage` table is maintained until session changes(close program, exit to main menu, change planets).  IIRC even /reload doesn't change the session anymore.

Only `storage.state` is kept between sessions.  Even that will be reduced to a boolean value.  So trying to make `storage.state` a table will only leave you with `true` upon reloading Starbound.  But that's fine if you only need a boolean value.

`storage.state = storage.state or false`
In case that looks funny, I'll explain.
if `storage.state` was `true` before, it still will be.  If `storage.state` was `false` or not defined aka `nil` then it will be assigned `false`

This handy way to keep the value is possible thanks to how LUA handles `or`
Any value not being either `false` or `nil` will evaluate as `true` so `or` will not check what is on the right.  But if `false` or `nil`, it will return the result of the right hand expression instead and therefore the return of that expression will be used for the assignment of `storage.state`

NotMrFlibble:
Looks somewhat more reasonable now, but I'd still do a squash-merge and I'd do it locally rather than via the Github web interface.

PR #422, opened 2016-10-29
Update isn_sharedobjectscripts.lua

bk3000:
Some efficiency improvement

### DIFF
--- a/objects/isn_sharedobjectscripts.lua
+++ b/objects/isn_sharedobjectscripts.lua
@@ -31,13 +31,9 @@ function isn_getAllDevicesConnectedOnNode(node,direction)
 end
 
 function isn_numericRange(number,minimum,maximum)
-	if number > maximum then 
-    return maximum
-	elseif number < minimum then 
-    return minimum
-  else
-    return number
-  end
+	if number > maximum then return maximum end
+	if number < minimum then return minimum end
+	return number
 end
 
 function isn_countDevicesConnectedOnOutboundNode(node)
--- a/objects/isn_sharedobjectscripts.lua
+++ b/objects/isn_sharedobjectscripts.lua
@@ -8,13 +8,13 @@ function isn_getAllDevicesConnectedOnNode(node,direction)
 	else nodeID = object.getInputNodeIds(node) end
 	if nodeID == nil then
 		---sb.logInfo("GGDOND: Node with ID " .. node .. " does not exist, terminating with nil")
-		return nil
+		return nil  --possibly you might just want to return {} instead of nil
 	end
 	
 	---local counter = 0
 	local devices = { }
 	---sb.logInfo("GGDOND: Iterating connected devices")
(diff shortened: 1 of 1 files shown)


bk3000:
isn_getAllDevicesConnectedOnNode() maybe should be returning an empty table aka {} instead of nil when nothing found but I didn't change this.
#{} would return 0 so you'd know it is empty and feeding pairs() or ipairs() and empty list will not execute the code block as there are no contents to the table
This will not have undesirable effects nor trigger an error and is therefore probably preferable.

NotMrFlibble:
Returning `nil` is fine. If you have it return `{}`, there may be `if` conditions to change too.

Perhaps you should look at all uses of `isn_getListLength(arg)` and directly replace with `#arg` where practical. If all uses are replaced, remove the function; otherwise keep it _unchanged_.

objects/isn_sharedobjectscripts.lua
```diff
@@ -31,18 +31,18 @@ function isn_getAllDevicesConnectedOnNode(node,direction)
...
-	if number > maximum then return maximum end
-	if number < minimum then return minimum end
-	return number
+	if number > maximum then 
+    return maximum
+	elseif number < minimum then 
+    return minimum
+  else
+    return number
+  end
```
NotMrFlibble: This function is concise and readable and (probably) optimal as is. The only maybe-useful changes here would be to use built-in functions.

NotMrFlibble:
I strongly suggest pulling all of the pairs()→ipairs() changes into one commit (and being sure that there are no cases where the table is not a list).

objects/isn_sharedobjectscripts.lua
bk3000: `isn_getListLength` is intended to get the list length.  `#`gets the list length.  I wasn't going to mention this but actually there is a difference which makes `#arg` better where outcome is concerned.  The existing function is flawed.  Your current usage doesn't trigger the flaw, but for example lets say someone did this

```
local myTable = {12321, 123123, 3.14, "string", {"Chuckle Sushi"} }
myTable.not_a_number = true
local mySize = isn_getListLength(myTable)
local i = 1
while (i <= mySize) do
  sb.logInfo( tostring(myTable[i]) )
  i = i + 1
end
```

That's all very typical code.  Here `isn_getListLength` will return `6` instead of `5`.  You'd end up referencing a nil index because `pairs()` will use non-numeric indexes as well as numeric.  For some things that's desirable but getting a list length isn't one of them.  You could change that to` ipairs()` to correct this but `#args` is just better.

Now this is only currently used in _objects\power\isn_radiostation\isn_radiostation.lua_ and only on an iterated list (I checked with notepad++'s "find in files") and as long it is never used any other way... I guess you'll be okay.

I do not understand the reason for your objection to changing it.  If you feel like I'm stepping on your feet... I'm sorry.  Just trying to be helpful.

PR #460, opened 2016-12-19
make refinery recipes output 50% tooltip price

UltraHex:
as per vanilla. ff_mercury has no tooltip value so leave ff_cinnabar.recipe unchanged.


UltraHex:
Is there any reason this isn't already the case? If there is, perhaps FU should nerf the vanilla recipes.

PR #502, opened 2017-01-24
add catchall

IsaacHart:
I dont understand why the various ranged weapons don't need to be included here.


sayterdarkwynd:
because they are included in gunfire.lua and bowshot.lua, because they need to use a different method to apply it

IsaacHart:
sorry for the confusion. the comment was unrelated to the utility of the catchall.

PR #509, opened 2017-02-02
Update and rename largeflying.monstertype to largeflying.monstertype.…

IsaacHart:
…patch


IsaacHart:
i should not be gitting drunk. apologies.

PR #521, opened 2017-02-09
Kherae s workbench

Kherae:
grav, antigrav, atmoreg and regenmatrix update.
and update for zero-g. *snickers*


Kherae:
dont take this til I get online. I missed a variable on asteroid field gravity :|

PR #566, opened 2017-03-02
Hot Fix

Tamamo89:


Tamamo89:
Fixed a specific wand ability being the wrong class.

PR #584, opened 2017-03-06
Lab Directory Fix

Karthas077:
Fixed Lab Directory blank arrow bug.


Karthas077:
Also made the Advanced Xeno Lab count as Tier 3. Duplicated Tier 2 times into Tier 3 so performance remains unchanged.

PR #647, opened 2017-04-10
Swap clonelab inputs to logical sides + typo fixes

Chicken-Nugget:
Cloning lab does the opposite of what would first thought usage would be. Sapling icons have trunk left and leaves right, so would think to put them in the cloning lab the same way (basically sticking them together), but it is opposing to this logic.

Also should not be printable.


Chicken-Nugget:
Also added some simply 'a/an' typo fixes in item descriptions

sayterdarkwynd:
seems a rather needless chang...e

Chicken-Nugget:
Yeah, nothing major, just that I'm often doing it the other way, guessing others are as well.

I know proper english is "An Avali", thats exactly what those description changes fix. As well as some instances of 'an' being used improperly.

Great mod btw, you have a ton of code here. Thought I'd help with some minor things like typos that you probably don't have time to worry about :)

Chicken-Nugget:
Finished scanning though all ~4,200 item descriptions, corrected about 50 typos.

sayterdarkwynd:
my bad. I think I looked at them backwards ...too early in the morning maybe when I looked last :) thanks.

PR #657, opened 2017-04-16
Fixed Chromatic biome incorrect effect

dualinfinities:
is this the right effect? also am I doing this right?


sayterdarkwynd:
looks good. nice catch. i hadnt even noticed.

PR #680, opened 2017-04-27
Distortion Snakes!

myothernameisreal:
New directory for segmented monsters.  Some improvements to monstersegment.lua that allow for monsters with multiple types of segments in one body.  Rendering layers are from head to tail now.  Removed whiplash effect.  Combined script for follow behavior with script for rotation frame by angular velocity for new visual effect.  They're still pushovers due to low HP but I added electrical resistance which I think will be necessary for all multi-monster entities.

Next challenge will be setting rotation by relation to parent segment or velocity which would be useful for more organic looking worm like monsters.


myothernameisreal:
The new monsters are
spheresnake
aquasnake
sonicsnake
spikesnake
if you want to get a quick look

PR #716, opened 2017-05-14
So many burrowers update

myothernameisreal:
Animations are properly assigned and no more asset exceptions should be thrown into the log.  All the segment animations are in line with their head counterpart for death effects.  Animations are appropriately lighted for ores with glow effects.


myothernameisreal:
Every elemental resistance profile has been reworked.  Some tuning may be in order eventually.  In general segments are very resistant to damage, heads somewhat resistant, and tails vulnerable.  Hopefully this changes the dynamic a bit from just trying to hit as many at once as you can.

myothernameisreal:
Monstersegment.lua changes seem to have resulted in status effect immune creatures.  Not sure how though.  Needs further testing, but the enemies are still defeatable without inflicting statuses.

PR #766, opened 2017-06-22
completes the blanaced explorer and protection armor collections

sam-tunder:


sam-tunder:
this now also fixes the small bug with the liquid collection

PR #774, opened 2017-06-25
Compress PNGs by means of PNGGauntlet

dualinfinities:


dualinfinities:
whoops, I put it the wrong way when trying to merge the new stuff you added. w/e though, still works.

PR #784, opened 2017-06-30
Small set bonuses description fixes

gabriellett:
Some pieces from the Sanguine armor set and the Black Slime armor set had different/wrong/missing stuff from their bonuses descriptions. This PR fixes it.

I also noticed that the "Dweller Armor" chest piece is missing the "Detects Ore(wip)", and I could not find the effect implementation. Should the "Detects Ore(wip)" be present on the chest piece or removed from all others?


sayterdarkwynd:
detects ore shoudl not be there at all anymore, no. that effect was too laggy and removed.

gabriellett:
@sayterdarkwynd ok! Removed detect ore thing from the head and legs pieces.

PR #803, opened 2017-07-08
completes the axe collection and also fixes a few bugs with katanas

shadow-absorber:


shadow-absorber:
someone made some mistakes when creating the recipes for the katanas so i fixed them

PR #836, opened 2017-07-21
Add massive worlds to all non-ocean planet types

jtgibson01:
By creating an additional "NoOcean" planetaryType for each orbital zone, it is possible to have the best of both worlds (*snerk*) by having huge, etc. worlds generate, but also have oceanic-type worlds restricted to no greater than large.  (My list: tidewater, ocean, strangesea, bloodstonesea, aethersea; may need refinement as I'm a Tier-3-for-lifer kind of guy and haven't witnessed the latter three.)

- Duplicates all planetaryTypes with an additional *NoOcean type
- Removes "tidewater", "ocean", "strangesea", "bloodstonesea", "aethersea" from *NoOcean planetaryTypes' terrestrialTypes list
- Adds additional *NoOcean planetaryTypes to orbitRegions of all systemTypes
- Weighting 0.4 Normal, 0.3 NoOcean, to preserve same generation frequency of gas giants and asteroids
- As NoOcean types can also generate smaller worlds, massive worlds are now about 20%-30% as common as pre-removal and even more awesome and special when they do occur
- They're great again!

Disadvantages:
- If a new world type is introduced, author will have to remember to add to both the normal and NoOcean terrestrialType list for appropriate system types
- Ocean worlds are slightly reduced in frequency (roughly 40%), although this is an advantage in disguise (we've got one known ocean-covered world in the known universe so far, and it's only 2/3 ocean, most of that only a couple kilometres deep)

### DIFF
--- a/celestial.config.patch
+++ b/celestial.config.patch
@@ -1429,7 +1429,8 @@
           "orbitRange" : [4, 5],
           "bodyProbability" : 0.7,
           "planetaryTypes" : [
-            {"weight" : 0.7, "item" : "binarySunScorching"},
+            {"weight" : 0.4, "item" : "binarySunScorching"},
+            {"weight" : 0.3, "item" : "binarySunScorchingNoOcean"},
             {"weight" : 0.07, "item" : "fugasplanet"},
             {"weight" : 0.04, "item" : "AsteroidField"}
           ],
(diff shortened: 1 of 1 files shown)


sayterdarkwynd:
not sure i want to implement this. I considered it, but think it might actually be better to avoid the huge planets save for fringe scenarios. It feels more authentic, and encourages more exploration. I'll have to think on this.

shadow-absorber:
hmmm this seems like a viable solution to me tough

PR #845, opened 2017-07-31
Temp fix for solar power on custom space stations

Gaikotsu2:


sayterdarkwynd:
appreciated, but pointless right now. We are working on a new power system.

Gaikotsu2:
Fair enough. Is the new power system able to be looked at? I would be curious how you are planning to improve it.

PR #897, opened 2017-08-11
Kherae s workbench

Kherae:
pumps and quarries!


sayterdarkwynd:
will wait till you are done with this.

trailing comma in the json file will result in breaky breaky object.

also, are you starting to add in the tiered pumps etc from asana that have sat their for months? would be great!

PR #915, opened 2017-08-22
More compatible Matter Manipulator solution

gamrguy:
No longer edits vanilla upgrade UI code, instead creating the new items with the proper parameters to be recognized properly. Solves the following issues:
- Matter Manipulators causing upgrade resets
- Still being able to upgrade non-vanilla Matter Manipulators
- Being stuck with a low-range Matter Manipulator if the upgrades weren't bought (they can be bought properly now; these upgrades use a player status, not an item parameter)
- Incompatibility issues with Matter Manipulator Manipulator (custom MMs now have their proper stats and can change sizes, toggle liquid collection, etc. just like a normal MM)

If this pull request is accepted, please make sure to change FR to allow upgrading of FU MMs!


sayterdarkwynd:
ive manually added this in, because for some reason it refuses to stop claiming there is a conflict even when i duplicate the files to their status prior to your changes. *shrug*

PR #928, opened 2017-08-31
liquid interaction fixes.

shadow-absorber:
but maybe not all of them. hard to tell but atleast it doesnt break old stuff as far as i can tell


sayterdarkwynd:
my penis is gigantic

PR #955, opened 2017-09-12
patches for vanilla backpacks

adaxe:
add

  {
    "op": "replace",
    "path": "/tooltipKind",
    "value": "baseaugment"
  },

to every existed vanilla pack patches in FU. They should now have augment slots.


sayterdarkwynd:
nice. thanks good sir!

PR #964, opened 2017-09-16
Moved some recipes.

Tamamo89:
All Leather and Bone is now on the Armory1 station.
All Iron, Tungsten, Titanium, and Durasteel are on their respect tiered anvils. Including shields.
All other vanilla ores are left untouched.
The Prismatic Staff and Wand actually unlock by Prisilite Stars now.


sayterdarkwynd:
Bone and Leather were intentionally on the Anvil and Crafting Wheel. Undo those changes. The rest are fine  . 

I have Leather and Bone on the other stations because they fit better there thematically, and progression-wise. By the time you get an armory, bone is more or less pointless anyhow since you can simply build better armor by then. Thus, at the anvil it gets noticed more because its one of the few things you CAN craft, and before you get enough metal for iron sets (generally speaking)

Tamamo89:
The main reason i moved the bone stuff is because some of it was on one station or the other, some of it was on both. If you want it on both, we'll just have to change it so all bone stuff is unlocked at the anvil. And no bone armor isn't pointless at all. You forgot about the upgrades. ;)

sayterdarkwynd:
I wanted them exactly where they were for good reason. I do not want the bone changes or leather changes made.

PR #968, opened 2017-09-19
longswords upgradeable

adaxe:
The random generated longswords seem missing "upgradeableWeapon" tag. Please ignore this pull request if they are intended not to be upgradeable.


Tamamo89:
Nope they aren't

shadow-absorber:
thanks for the fix. really something i should have taught about when we added those.

PR #970, opened 2017-09-20
Update 8bitpowerup.augment

Slugbaitthenotreallycoder:
Why does /n not work? I have no idea. however, just moving shit down a line does


shadow-absorber:
well normal escape sequences dictates that you use \ not /
so it should be \n 
@Slugbaitthenotreallycoder

PR #972, opened 2017-09-25
Miscellaneous Fixes

Hubnester:
Fixes the blast furnace and arc smelter using protorock instead of protorockmaterial and removes the information in the description of the weapon turrets about them not being able to be picked back up when they can be.


sayterdarkwynd:
already manually applied these from your zips :) glad to see you on git though. will make it easier.

PR #980, opened 2017-09-27
Horse death sound

Hubnester:
Changes the sound of a horses death to the death sound from the zip file (couldn't get the idle and damage sounds working but while I was investigating it appears there is currently nothing that uses those or has those working)


sayterdarkwynd:
idle sounds and shit are handled in the animation file instead.

take a peek at how some of the FU mosnters work. i suggest  /boss/gorillaking, as he has numerous sound triggers in his animation file.

PR #1003, opened 2017-10-03
Tech crafting + collection changes

Hubnester:
Changes the tech collection to use the tech icons for it's icons and makes tech crafting directly unlock the recipe unlocks and collectable (no longer gives the item)


shadow-absorber:
@Hubnester thank you for doing this.

PR #1011, opened 2017-10-05
Briefcase O' Bees

Hubnester:
Adds a briefcase o' bees that you throw and it will explode into 10 random bees (1 in 20 chance of being a naturally spawning apiary type bee and the rest are aggressive bees (only the ones in the bees spawntype file). Also includes a version for Kevin to give (if you decide to) that acts like a reward bag instead of a projectile. It's currently unobtainable and probably needs tweaking (the bees seem to do too much damage on a tier 1 planet (scales with planet tier)) and I have no idea what else to do with it.


sayterdarkwynd:
cool. lets add it to the random loot pool with Crunchy Chicks and the like   :)

PR #1014, opened 2017-10-05
Changed shortDescription of isn_battery_t3_ceru to show correct (8W) output

Manlibear:
Was showing `6W` before, but in the .object it says `"isn_batteryVoltage" : 8`
As a stylistic thing, would you prefer it non-prettified?


sayterdarkwynd:
already fixed on my end

PR #1035, opened 2017-10-14
Music Player Improvements

Hubnester:
Music player is better (breaks old patch files for it)


Hubnester:
Also made a PR for the new patch file for FM

PR #1054, opened 2017-10-21
Fixes microsphere techs oxygen issue

Hubnester:
Fixes the player suffocating when using any of the microspheres when there is a block directly above you. (NOTE: microshpherespider and mircosphereprecursor now use a seperate lua then the vanilla distortion sphere lua)


sayterdarkwynd:
they...already did use their own lua files?

PR #1062, opened 2017-10-24
Liquid Mixer and Liquid Interaction Changes

Hubnester:
Adds material + liquid interaction recipes to the liquid mixer. Fixes a couple blocks that interact with liquids and removes duplicate liquid interactions (removed the interaction that doesn't work). REMOVED A FEW INTERACTIONS THAT GIVE ELDER FLUID!!!


Hubnester:
btw the elder fluid info in caps is because I've noticed you sometimes copy what I put here for the change notes about it and I wanted to make it more obvious to people that some of the interactions were removed

PR #1064, opened 2017-10-25
Stops spam from beer epp when inventory is full

Hubnester:
Makes it so that the beer epp won't drop liquidbeer if there is a full amount on the ground in a 5 block radius (to prevent massive buildups of it if the player has a full inventory)


Hubnester:
Might not actually work (and has been improved)

PR #1075, opened 2017-10-28
Atmospheric Filter

Hubnester:
Adds atmospheric filter (has a lot that an augment can be put to apply it's effect to all players and npcs in range), changed the atmospheric regulator and regeneration matrix to use the same lua (has 0 slots for augments currently). Made it so that the liquid mixer recipe is unlocked by interacting with an electronics center or power station.


Hubnester:
btw the current range of the atmospheric filter is 500 blocks (the same as the atmospheric regulator and regeneration matrix)

PR #1080, opened 2017-10-29
Status Tablet Image Fix (hopefully temp)

Hubnester:
Copies statustablet.png and renamed it to techtablet.png (to temporarily fix the issue until the cause is discovered)


Hubnester:
Ignore this PR if you already know what the issue is

PR #1107, opened 2017-11-09
Science Outpost Bar Improvements

Hubnester:
Make's it so that the bar spot no longer renders on top of the player, makes it have platform collision on top of the table part and makes the starbooze npc use lounger behaviour.


Hubnester:
Also modifies the lounger behaviour to only have a range of 1 (all the vanilla npcs that i know of that use this behaviour seem to still work fine)

PR #1123, opened 2017-11-14
Added "Mech Aim Actuator"

kittycat2002:


whismerhill:
isn't it lacking something to unlock it right now ? though I guess having it non available at the start is a good idea to test it out ?

PR #1129, opened 2017-11-17
Ship Speed Modification

Hubnester:
Allows ship speed to be modifed. Adds small thrusters that increase ship speed slightly. FTL Drive ship speed increase now works (you will need to pick it up and place it again for it to apply) + you can only place one of them.


Hubnester:
Will cause issues with already placed FTL Drives (will not add the ship speed bonus + will remove it even though it isn't applied)

PR #1148, opened 2017-11-23
Minor Fixes

Hubnester:
Fixes a painting having the wrong frame size.
Makes it so that the BYOS FTL drive quest requires a small FTL drive instead of a normal one.


sayterdarkwynd:
already made both the same changes on my end lol

PR #1173, opened 2017-12-14
Simplified processor recipes

yerinn:


sayterdarkwynd:
nope. this makes them far too cheap.

sayterdarkwynd:
am open to suggestions, but this just whitewashes them too much.

sayterdarkwynd:
however...using this as a base i think i do have a solution. its minor. will append and commit.

PR #1211, opened 2018-01-14
Not ready for merge IMO, more interested in feedback...

ProfoundDarkness:
overview of changes
-ONLY changes the lua for fu_growingtray and not (yet) isn_hydroponicstray
-Instead of FU having to know the seed output, look up the treasurePool from the seed.
-Base the grow time on the actual seed.
-Instead of one item type from the output yield multiple types based on the treasure pool.
-Currently does multiple iterations of the treasure pool (fertilizer).
-Only uses 1 unit of water per growth stage.

Details:
-init-
insert seed
if seed passes checks (original plus a check if it is a farmable type)
-copy information from seed .object definition, specifically "stages" into seedData.stages
-Also fetch the seed treasure pool ("harvestPool") and if the seed is perennial/annual
-generate growthData from seedData.
--Basically how long to grow as well as how long (of total grow time) each stage would be.
*might need to insert a flag to consome one water as currently that might be free for the first stage.
-check fertilizer state and set yield count

-growth-
Check if we are in an apt state to update...
Calculate which graphic (from 0 to 3) to show based on how far the plant has grown.
Check if a new growth stage was entered...
-If so then try to use some water.
 *pretty sure that the first phase is free for water
--If the water is used, keep growing and increment the internal stage to track when the next water happens.
--If not then growth is haulted until water is present.
Update growth progress (considering fertilizer)
Check if growth complete...
-If growth complete then for every yield count...
--generate contents of treasure pool (a list of item descriptors)
---for each of those add them to the container (same way as before).
--determine if the plant is perennial or annual...
---if perennial then simply reset (regenerate) our growth data to fit the reduced number of stages and resume...
---if annual then much like the old mechanic of consuming another seed.

Open Questions:
-The right way to handle 'water'?  Should it constantly use like before or use per growth.
-Should the last growth stage be used?  I have a strong impression that the final stage is the old
 rot stage and should be ignored... It's usually a VERY large number (thousands) where other stages are hundreds or tens.
-Should there be something to speed up the growth more?
--Higher growth values per tick?
--Division of growth stage (ie if total growth is 2000 then divide that by 5 or something).
-(yield)Could generate multiples of a single treasure pool instead of multiple treasure pool iterations.

TODO:
Besides filling in some blanks and testing... really need to clean up this mess as there was a LOT of learning going on...
Verify fertilizer mechanics are still working.


ProfoundDarkness:
Did some more research and managed to answer one of my questions... the final stage on the crop is NOT to be used to determine how long the crop grows, previous stages yes, final no.
I'll make a tweak soon to fix that as right now they take *WAY* too long to grow.

sayterdarkwynd:
id like to see water cost configurable based on a config param so that fertilizers can impact amount of water required per phase. That's the ideal, there.

Fertilizer speeds up the rates currently, and should do that in your revision as well. again, see params with settings i can change per fertilizer, etc.


i like this on paper so far

ProfoundDarkness:
So there are now config params (that are used) for the following things.
-Base growth rate (default is 1)
-Liquid items not only have a fluid value but also a growth boost value (since fluid is far less granular now).
-Fertilizer items have 3 values, how much fluid use they cause (this is not additional fluid use), additional growth speed, and additional harvestPool picks (iterations).

Plants only use one fluid (water) per growth stage (including initial stage).  They consume a fertilizer at the start of growing (so if inserted later has no effect until the next growth cycle).

Question: storage.activeConsumption is related to powered items isn't it?  Does it fit on this unpowered item?  The growing tray is an unpowered item right?

ProfoundDarkness:
Basically awaiting feedback now.  I'll be doing some further testing and bug hunting/fixing but without feedback it's not going to change much from here.
I've got a bit more init state checking to do too.

ProfoundDarkness:
K... gonna close this exploratory request, it was never really meant for merging.  Will re-submit a new PR after I've moved the isn_hydroponicstray over to this mechanic as well.

PR #1220, opened 2018-01-17
Branch trays use seed data

ProfoundDarkness:
Ran with the ideas I presented earlier.
Trays (fu_growingtray and isn_hydroponicstray) use the same lua file so changes to one alter the other.

Basically both trays now act more like plants between growth time, water use, seed use, outputs.  You can use fertilizer and liquids to alter the pace of growth and the object files can setup how different things work (including base growth rate).

Overview of changes:

The object files have several configurable parameters.
-Can set how fast the tray grows seeds at.
-Can set if the tray handles saplings.
-Water given from liquid items is not only how much water it's worth but a growth boost.
-Fertilizer can impact water use (even 0 use), growth speed, and number of picks when grown.
-Removed the old table of seed->output as it's not necessary.

In the code the trays use actual seed data for growth time, frequency of watering, and for treasurePool or amount/type of wood/saplings from trees.

Plants are watered once per growth stage, plants are expected to have 2 growth stages.

Growing Tray basically uses Hydroponic Tray growth code (reminder the object file can set growth).  That means that Growing Tray grows the plants per second instead of per update.
Biggest change to Hydroponics tray (besides the above) is it will no longer try to use multiple seeds in one go.

Oh and breaking the tray recovers the currently growing seed.  The seed is never fully gone until growth is complete.


ProfoundDarkness:
Found a bug in perennial plants, turns out the final growth phase IS used, reading strongly suggested it wasn't but several tests show it is.

ProfoundDarkness:
Tested with a modified plant which had one fill and one 100% chance pool of 1 item with 100% chance to pick one from the pool.  Multiple tray and multiple plant (in dirt) outputs matched so confidence is very high that outputs from trays match if the seed were planted in dirt (whatever seed type).

More testing of the new code paths needed (ie play-testing tomorrow).

sayterdarkwynd:
i havent had a moment to review the code yet. 

"Biggest change to Hydroponics tray (besides the above) is it will no longer try to use multiple seeds in one go."

what? no. that contradicts the entire rewrite i literally just finished for all the fertilizer types. The whole point was that both machines required more than a single seed, and it could be mitigated with fertilizers, making it less a fire/forget scenario unless you are willing to spend time on more elaborate fertilizers.

unless i am misunderstanding? 

the rest seems sensible enough.

ProfoundDarkness:
... just realised... I was so focused in on how the old growing tray code worked... the graphics show 4 plants growing.

ProfoundDarkness:
I might be totally off base with my balance settings... however I'm more interested in getting the code into a state that is desired... Balance is rather beyond me due to inexperience with the mod.

sayterdarkwynd:
balance, we can worry about :) im more concerned about the featureset

and yea, multiple seeds are used because it has "several" plants. i went with 3 default rather than 4 to ease it a bit

sayterdarkwynd:
I think we're basically on board with this. We're poking at it a bit before working it in, to suit recent adjustments and add a bit more mutability. But, I like! well done good sir.

ProfoundDarkness:
I finally got around to testing the hydroponics tray (I haven't gotten near power in my main game yet so that was a lot of reading).  There were some glaring bugs in the power mechanics which should be resolved now.

Also having done some tests and reading to learn (a little) about how power works I modified the Fast Forward mechanic so that it will try to do smaller steps (up to a point).

This next one really could use some guidance...
I see 4 ways to handle the power Fast Forward situation (basically naive approaches below).
-1 - The way I already did, basically do fast forward in smaller stages, up to a point (default is 1 minute increments, requires 120 power).
-2 - A loop which tries to use the normal amount of power rapidly in small chunks.  That one will weigh on a server some.  Granted the loop would only run for like 60 cycles per update.  Based on the defaults (configurable at the top of the lua file).
-3 - The f**k it approach, just fast forward and if you've got the power to run the machine in that state, yay, if not, then have a bit under half speed growth for like a minute for the hour you were away and be happy you got that.
-4 - Basically the device uses the normal power but you get the fast forward results... this one feels a tad more cheat-y to me.

Also the hydroponics tray is way out of balance (growing time, slow) right now considering I didn't notice it used a different delta time than growing tray, but that's just a tweak in the object file, don't need to touch code.

sayterdarkwynd:
okay so a suggestion
replace
function cRequiredPower()
     return config.getParameter("isn_requiredPower") or 0
 end

with something like
self.requiredPower = config.getParameter("isn_requiredPower") or 0

in the init function

whismerhill:
@ProfoundDarkness just no loop, we should leave scripts as lightweight as possible. as huge bases already have performance issues in starbound.
hence solution 2 is a big no as far as I'm concerned

whismerhill:
reformatting sayter's copy/paste for better readability : 
```lua
if secondsThisUpdate > maxFastForwardSeconds then secondsThisUpdate = maxFastForwardSeconds end
```
should be
```lua
secondsThisUpdate = math.min(secondsThisUpdate,maxFastForwardSeconds)
```
and replace
```lua
function cRequiredPower()
     return config.getParameter("isn_requiredPower") or 0
 end
```
with something like
```lua
self.requiredPower = config.getParameter("isn_requiredPower") or 0```

File: objects/isn_unifiedgrowingtray.lua [hunk 1]
```diff
@@ -0,0 +1,524 @@
...
+	elseif growthperc >= 50 then
+		animator.setAnimationState("growth", "2")
+	elseif growthperc >= 25 then
+		animator.setAnimationState("growth", "1")
+	else
+		animator.setAnimationState("growth", "0")
+	end
+	
+	--Check if a previous fluid use failed...
+	if storage.hasFluid == false then storage.hasFluid = isn_doFluidConsume() end
```
whismerhill:
```lua
if not storage.hasFluid then
	storage.hasFluid = isn_doFluidConsume() 
	if not storage.hasFluid then return end	--Failed again?  No growth nor outputs.
end
```
no need to check twice if storage.hasfluid is true from the start

whismerhill:
pardon me for the multiple comments, so many changes that I'm a bit lost.
So ...  the new tray's code gets outputs directly from config files of plants instead of having its own separate tables, right ???? that's an improvement in my book

objects/isn_unifiedgrowingtray.lua
```diff
@@ -0,0 +1,524 @@
...
+		local stemPath = root.treeStemDirectory(stemType)
+		local stemJson = root.assetJson(stemPath .. stemType .. ".modularstem")
+		storage.stemDrops = stemJson.dropConfig.drops
+		storage.hasTree = true
+	end
+	if foliageType then
+		local foliagePath = root.treeFoliageDirectory(foliageType)
+		local foliageJson = root.assetJson(foliagePath .. foliageType .. ".modularfoliage")
+		storage.foliageDrops = foliageJson.dropConfig.drops
+		storage.hasTree = true
```
whismerhill: I'm pretty sure a valid sapling HAS to have a valid stemName.
I'm not 100% sure but back then when the stemName was invalid it resulted either in Generic saplings or PGIs (Perfectly Generic Items, e.g. a failure in essence)

whismerhill: hubnester & me just talked about it, we both think both should be valid at the same time

objects/isn_unifiedgrowingtray.lua
```diff
@@ -0,0 +1,524 @@
...
+			if storage.growth < storage.stage[storage.currentStage].val then break end
+			if storage.hasFluid == false then break end
+			storage.hasFluid = isn_doFluidConsume()
+			storage.currentStage = storage.currentStage + 1
+		end
+	end
+	
+	--If the above fluid use left us dry stop here (don't output items) and try again later.
+	--Yes this may generate 1 more (unnecessary) growth spurt but it won't use more fluid.
+	if storage.hasFluid ~= true then return end
```
whismerhill: `if not storage.hasFluid then return end`
unless you have a specific reason to check for true specifically ???

objects/isn_unifiedgrowingtray.lua
```diff
@@ -0,0 +1,524 @@
...
+	
+	--Check if we are grown.
+	if storage.growth >= storage.growthCap then
+		--Spawn our output...
+		for i = 1, storage.yield do
+			local itemDescriptors = nil
+			if storage.hasTree == true then
```
whismerhill: `if storage.hasTree then`

objects/isn_unifiedgrowingtray.lua
```diff
@@ -0,0 +1,524 @@
...
+				itemDescriptors = isn_genSaplingDrops()
+			elseif root.isTreasurePool(storage.harvestPool) then
+				itemDescriptors = root.createTreasure(storage.harvestPool, world.threatLevel())
+			end
+			if itemDescriptors ~= nil then
```
whismerhill: `if itemDescriptors then`

objects/isn_unifiedgrowingtray.lua
```diff
@@ -0,0 +1,524 @@
...
+	end
+	
+end
+
+--Updates internal fluid levels and modifies a growth variable depending on
+--liquid in slot.
+function isn_doWaterIntake()
+	local contents = world.containerItems(entity.id())
+	local water = contents[storage.waterslot]
+	if water == nil then return false end
```
whismerhill: `if not water then return false end`

objects/isn_unifiedgrowingtray.lua
```diff
@@ -0,0 +1,524 @@
...
+	
+	for key, value in pairs(config.getParameter("isn_waterInputs")) do
+		if water.name == key and type(value) == "table" then
+			local amount, boost = value[1], value[2]
+			if amount == nil or boost == nil then
```
whismerhill: `if not amount or not boost then`

objects/isn_unifiedgrowingtray.lua
```diff
@@ -0,0 +1,524 @@
...
+	return false
+end
+
+--Attempts to use up some fluid
+function isn_doFluidConsume()
+	if storage.fluid < storage.fluidUse then
+		--Not enough internal fluid to keep growing, attempt to consume an item.
+		--May need to consume multiple items to get fluid level high enough.
+		while (storage.fluid < storage.fluidUse) do
+			if isn_doWaterIntake() == false then return false end
```
whismerhill: `if not isn_doWaterIntake() then return false end`

objects/isn_unifiedgrowingtray.lua
```diff
@@ -0,0 +1,524 @@
...
+	--Consume some fluid
+	storage.fluid = storage.fluid - storage.fluidUse
+	return true
+end
+
+--Fetch the seed from the storage slot, also does validity check.
+function isn_readContainerSeed()
+	local contents = world.containerItems(entity.id())
+	local seed = contents[storage.seedslot]
+	if seed == nil then return nil end
```
whismerhill: `if not seed then return end`

objects/isn_unifiedgrowingtray.lua
```diff
@@ -0,0 +1,524 @@
...
+	local seedConfig = root.itemConfig(seed).config
+	if seedConfig.objectType ~= "farmable" then return nil end
+	if seedConfig.objectName == "sapling" and cHandlesSaplings() ~= true then return nil end
+	return seed
+end
+
+--Reads the currentseed's data.  Return false if there was a problem with the
+--seed/data.
+function isn_readSeedData()
+	if storage.currentseed == nil then return false end
```
whismerhill: `if not storage.currentseed then return false end`

objects/isn_unifiedgrowingtray.lua
```diff
@@ -0,0 +1,524 @@
...
+	storage.resetStage = 0
+	storage.stage = {}
+	storage.harvestPool = nil
+	--Tree data
+	storage.hasTree = false
+	storage.stemDrops = nil
+	storage.foliageDrops = nil
+	--pairs() has an undefined order...
+	local index = 1
+	while seedConfig.stages[index] ~= nil do
```
whismerhill: while seedConfig.stages[index] do

objects/isn_unifiedgrowingtray.lua
```diff
@@ -0,0 +1,524 @@
...
+	return config.getParameter("isn_defaultWaterUse") or 4
+end
+function cUnpoweredGrowthRate()
+	return config.getParameter("isn_unpoweredGrowthRate") or 0.434782609
+end
+
+--Returns active seed when tray is removed from world, much like how plants
+--work.
+function die()
+	if storage.currentseed ~= nil then
```
whismerhill: `if not storage.currentseed then`

whismerhill:
I think you should have gotten the trend with all these already :p 
there's some more to edit, so good luck with that.
I'll get back to reviewing once you're done if you don't mind :)

objects/isn_unifiedgrowingtray.lua
ProfoundDarkness: OK, this one is a bit fuzzy to me.  I get nil == false in the boolean context (ie type) but if the type is non-boolean?  If you know of a handy reference (searching has failed) I'd love to read it.

I'm far more comfortable with the explicit nil check in this context.

Is the following (pseudo-code) a valid way of thinking of things in this context?
Taking just amount for example, is this basically saying (conceptually)
```
branchcheck(variable)
  if variable == nil then return false end
  if type(variable) ~= boolean then return true end
  --if type(variable) == boolean then return valueof(variable) end
  return valueof(variable)
end
```

Also if that is valid then I'd expect the following to also work:
```
if not (amount or boost) then
```

objects/isn_unifiedgrowingtray.lua
ProfoundDarkness: Yes the code reads and uses the actual treasurePool/drop tables so FU doesn't have to know about a seed/plant and it's output for trays to work with them.  That was the biggest thing that got me started on this.

It also aims to handle perennials properly, that is most perennials don't have many/any seeds in their drop pool and we are no longer explicitly dropping extra seeds so then (imo) it's better not to just keep eating seeds.  If the goal were to eat seeds it is easy to exclude seeds from the drop outputs.

I should note that a lot of the changes besides drop table/pool (ie timing of water use) fell out of having more data to work with and isn't really something I'm strongly for/against.  More did it because it was in front of me and I was aiming to model how plants grow in the game in dirt.

As for saplings, I believe you are quite correct regarding stems though it gets murky on foliage.  There are some trees which do not drop saplings and I don't know if the reason for that is they don't have foliage or their foliage drop table is empty.  Requiring both is sensible and can be revisited if a counter example (no foliage) is found.

ProfoundDarkness:
The last commit chiefly altered the middle portion of update().  Other changes in code were due to using the concept of self.variable in init instead of config reading functions.

"freePowerSeconds" is how much time since last update (ie fast forward) should only be charged 1 seconds worth of power.  Beyond that the hydroponics tray gets switched to unpowered mode for fast forwards.

Balance observation: maxFastForwardSeconds 1 hour seems a bit too small but 24 hours seems *WAY* too big.

I'd really rather just fully grow the crop each update (if we have that much stored time) but it's non-obvious how much 'water' to consume...

objects/isn_unifiedgrowingtray.lua
```diff
@@ -0,0 +1,496 @@
...
+	else
+		deltaTime=deltaTime+dt
+	end
+	
+	storage.activeConsumption = false
+	
+	--Try to start growing if data indicates we aren't.
+	if not (storage.currentseed and storage.harvestPool) then
+		storage.lastWorldTime = nil
+		if isn_doSeedIntake() ~= true then return end
```
whismerhill: `if not isn_doSeedIntake() then return end`

objects/isn_unifiedgrowingtray.lua
```diff
@@ -0,0 +1,511 @@
...
+	if initStage > storage.stages then
+		initStage = 1
+		storage.resetStage = 1
+	end
+	storage.currentStage = initStage
+	storage.growthCap = 0
+	local index = initStage
+	while index <= storage.stages do
+		local stage = storage.stage[index]
+		if type(stage) == nil then return false end
```
whismerhill: `if not type(stage) then return end`

whismerhill:
maxFastForwardSeconds 1h40minutes maybe ?
because : https://starbounder.org/Crops
most crops do grow in less than 94 minutes and regular crops will stop at just one stage of growth, hence I'd believe the maximum should be the same more or less ?

ProfoundDarkness:
I can't reply directly to the observation of the appearance of lots of calls in init() to isn_readSeedData() happening.

isn_readSeedData() reads the seed data for the current seed (not what is in storage but what is being grown).

What that whole chunk of code does is basically if something is wrong with the stored data (the cache) then read the data again.  If the data read failed then bail. Generally speaking the call will only happen once something is found to be invalid in storage and then the further checks won't fire since now there will be something valid (or something went VERY wrong).  Also this is in init() which is called once per entity (if I follow SB API correctly).

There is a loop in there that's not a check so much as to handle some strange behaviour I kept encountering where after saving and loading a game the keys in stage[key] were string versions (stage[tostring(key)]) of the integer keys and so that loop puts integer keys back IF that happened.

The last couple of times I tried to flag-ify that segment it added additional complexity that seemed quite pointless next to the pattern I've employed.  I'll look again at something when I'm more awake, maybe an additional call.

For example lets say that storage.stages (how many stages for the plant there are) is nil.  At that point we have data for a currentseed (we should be growing).  Lets say then that isn_readSeedData() fails, we blank the current seed (no longer growing) and return, no other calls or checks are performed.

Lets say the call succeeds instead, now the check below for if type(storage.stage) ~= "table" shouldn't end up calling again because we *just* read the seed data and stages should be valid (since the call succeeded).

Now lets say, in a fresh context, stages is valid but storage.stage (actual data about the plant's stage) is not a table, NOW we call readSeedData() and if that fails then we blank the current seed (stop growing) and don't check anything else.

objects/isn_unifiedgrowingtray.lua
```diff
@@ -0,0 +1,496 @@
...
+	storage.growth = storage.growth or 0 				--Plant growth completed
+	storage.growthFluid = storage.growthFluid or 0		--Growth boost from fluids
+	storage.growthFert = storage.growthFert or 0		--Growth boost from fertilizer
+	storage.fluid = storage.fluid or 0					--Stored fluid
+	storage.currentStage = storage.currentStage or 1	--Current plant stage
+	storage.resetStage = storage.resetStage or 0		--Stage to jump back to post harvest (perennial)
+	storage.hasFluid = storage.hasFluid or false		--If false soil is dry and no further growth.
+	storage.hasTree = storage.hasTree or false			--Used to track if seed data is sapling/tree.
+	
+	if storage.activeConsumption == nil then storage.activeConsumption = false end
```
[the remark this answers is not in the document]
ProfoundDarkness: That line was in there from the beginning, I've no clue what it does but I never changed the activeConsumption lines and don't know their purpose at all.  Made the change anyway.

objects/isn_unifiedgrowingtray.lua
ProfoundDarkness: Similar thing here... we are intending to return a boolean.
With what you are suggesting (for the return type) means that the following lines of code are no longer equivalent when they should be.
`if isn_genGrowthData() == false` then ... end`
`if not isn_genGrowthData() then ... end`
Proper behaviour in this case is since we are intending to return a boolean, explicitly return the false and not let it go nil.  This is because nil ~= false.

objects/isn_unifiedgrowingtray.lua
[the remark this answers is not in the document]
whismerhill: fine by me :)

ProfoundDarkness:
There may be a bug... can't investigate until tomorrow.

whismerhill:
the thing is : we don't want trays to grow stuff while you're away with zero effort
spoke with sayter about this earlier, he agreed with the current limit, we can always touch this stuff later anyways, not a huge matter.
As for the growth time, I'm not sure how the cycles & 83 seconds relate to plant growth time.
imho the base time should be related to how fast they grow as standard crops.
I should pick this stuff & try it out in SB directly.

about isn_readSeedData() seems like I misunderstood or misread.

ProfoundDarkness:
I don't know what effort is expected to be involved.  I've based the new code (sometimes somewhat loosely now) on what the old code did.  The old code would output results when a plant was grown, take in resources at the start and while growing, and grow plants while you were standing around.  The new code outputs results when a plant is grown, take in resources at the start and while growing, and grow plants while you are standing around as well as when you leave for a bit.  I'll give a summary of more specifics (old vs new and what to expect) later.

ProfoundDarkness:
I'm a bit confused at how fast forwards and some other changes are referred to as removing effort or seemingly free.  There isn't that much difference from an old growing tray's behaviour and the new growing trays behaviour (on annual seeds).  They both consume seeds, they both consume liquid items, they both consume fertilizer which tweaks various things, they both automatically harvest and replant.  There are some different details but the overall flow isn't different.

Fast forwards DO consume resources the same as if you were standing next to it.  It's NOT free, it's not cheaper.  I don't see standing next to something for a while as a form of effort and apparently neither did the Starbound developers or plants would still require you to hang out next to them to make any progress towards growth.

In general, because trays are so much like plants in soil I felt it apt to adopt as much as I could about how plants work into the tray's mechanics.  It helped having access to the real seed data so I had more points to hang interesting stuff from.  I've tried to expose what I can to configuration files (almost everything that is a number) for balance tweaking.  Almost everything in the object files is outside of my experience base so I can't really have a meaningful opinion on their settings, what is there is just there while developing, in a few cases like fertilizer I tried to pull data from code context into json context.

As for feature sets and their importance to me:
-Plants using real seed data for outputs, this THE big one for me and got me started on this.
-"Perennial behaviour"  Mainly the not using extra seeds when starting growth up again.  This is because the only guaranteed source of seeds now is from the seed's own drop table so perennial plants would be a bad idea in a growing tray without that.  The shorter growth time of additional growths is just because I could (that's the actual seed data).
-Fast forward feature (in some form) is of some importance to me for the same reason not having to stand next to a plant in dirt to get it to grow was important to Starbound players in the early alphas.
-Just about everything else I'm game for dropping/changing/keeping.  I am a *little* attached to what's there since I went to the trouble to code it but if behaviour is undesirable then so be it.  Generally speaking the reason it's that way right now is because that is how plants work in Starbound.

Going to put together a document on my fork (the wiki area) to have more details and information about settings.  That should be handy for those looking to balance this.  Also should help allow for meaningful discussion about further changes without having to understand/heavily play test the code.

When that's largely ready, will replace this line with a link.

whismerhill:
the thing is : standard soil planting only ever grows one, trays should behave more or less the same or else soil planting becomes useless which is not the intent. the zero effort was meant in this way.
e.g. don't go away then come back to 99999 products

As far as I'm aware(I'm not sayter), effort is meant as in like : make the effort of making fertilizers & actually expanding ressources to get more benefits. And unlike say.... plug to power, plug to infinite water ripe results which was really what was happening before.

Thanks for all this, your involvement, your time ....

ProfoundDarkness:
... I really wish I knew how to do sanity checks on load time instead of runtime... will be researching that.

Ah, OK.  My assumption was that because the growing tray's behaviour required no interaction from the user to continue through the stages/harvests that it shouldn't have to stop on any of the phases due to a player's temporary absence...  As long as there are resources to keep going that is.

Rather than a big long post of pros and cons of options gonna make a tweak on the code.  The current fast forward mechanic is flawed and it shouldn't be much to greatly improve on it.  (I hinted at it in the end of the wiki.)

whismerhill:
I don't think you need to do so many sanity checks in my humble opinion
since Lua is very lenient on variables types & whatnot
example : "10" + 1 = 11 
lua autoconverts the string to number
which means for example a defaultfluiduse of "3" is not a number & will be therefore be changed to the number 4 by your code, however "3" was just fine most probably

ProfoundDarkness:
When I did some testing (string compared to number) assuming leniency I got some script crashes so I went with checks like I did.  Lua is indeed lenient but only in certain contexts.  I'm not 'done' with init just yet, want to whittle it down some.  There are some sanity checks that absolutely can't go away like growth of 0 (tray sits and does nothing) and especially water use of 0 (can cause division by 0).

I'd really like to just have the checks done on object load rather than at runtime but I've still no clue if that can happen.

So regarding the fast forward mechanic.  I've modified the mechanic to allow the object to set how many fast forward updates are allowed in a row.  It's far less clunky and should be easier to pick a balanced value.  A growth cycle is interchangeably the plant finishing a current stage or a plant finishing growth to harvest.  The tray makes a decision about further fast forward updates when a plant's cycle completes (if completed fast forwards < max fast forwards, then more are allowed).  At least a value of 1 is suggested as 0 will truncate growth (even pre-harvest growth) which was payed for in resources even if the chunk has been loaded the whole time and the player is lingering nearby.

Here are some options moving forward with the mechanic:
- Remove it - I'm not a fan, some fast forwarding is desired, IMO, if for no other reason than plants do it.  The question that I'm wide open to is what degree/mechanism employed.
- Alter what 'cycle' means to ONLY relate to a plant moving from one stage to another.  Would simplify the code somewhat from line 176 to 194.  Currently a cycle is either 1 plant stage or multiple stages all the way to fully grown and ready for harvest - depends on growth state and how long it's been since the last update.
- Some other mechanism...

whismerhill:
1. keep the fast forward mechanic
2. about sanity checks we (modders) are gonna feed the correct data to it, if those types are wrong, there's already something very very wrong to start with. same with the water use, if the JSON contains a water use of 0 then it's fine if it crashes because the JSON would be incorrect.

objects/isn_unifiedgrowingtray.lua
```diff
@@ -0,0 +1,541 @@
...
+--Updates internal fluid levels and modifies a growth variable depending on
+--liquid in slot.
+--optional arg fluidNeed is amount of fluid required to top up to.
+function isn_doWaterIntake(fluidNeed)
+	local contents = world.containerItems(entity.id())
+	local water = contents[storage.waterslot]
+	if not water then return false end
+	
+	for key, value in pairs(config.getParameter("isn_waterInputs")) do
+		if water.name == key and type(value) == "table" then
```
whismerhill: you can replace those lines with something like : 
```lua
	local watertable = config.getParameter("isn_waterInputs")
	if watertable[water.name] and type(watertable[water.name]) == "table" then
```
notes: you need to change the variable name I came up with.
note2: this table need to be always loaded in init imho because we're not gonna reopen it each & everytime
note3: again too many sanity check, is there a valid reason why type(watertable[water.name]) (or type(value) == "table" in your original code) wouldn't be a table ??? & before you ask, no, modder error is not a valid reason.

ProfoundDarkness:
Please explain why I get the following error:
`[Error] Exception while calling script init: (LuaException) Error code 2, [string "/objects/isn_unifiedgrowingtray.lua"]:11: attempt to compare string with number`
When I change the following fu_growingtray.object code from:
`"isn_baseGrowthPerSecond" : 1,`
to
`"isn_baseGrowthPerSecond" : "1",`

AND I change the following lines lua lines:
`10	self.baseGPS = config.getParameter("isn_baseGrowthPerSecond")
 11	if type(self.baseGPS) ~= "number" or self.baseGPS <= 0 then self.baseGPS = 4 end`
`10	self.baseGPS = config.getParameter("isn_baseGrowthPerSecond") or 4
 11	if self.baseGPS <= 0 then self.baseGPS = 4 end`

objects/isn_unifiedgrowingtray.lua
ProfoundDarkness: RE the key lookup... <forehead> yeah, that was the structure of table lookups in the original code (both growingtable.lua and hydroponicstray.lua)... I didn't even think about changing it to the more efficient key lookup.  Good point on just caching the table too.  Only issue I see is it could increase the memory load some since EACH tray in the game will be carrying around a copy of those tables (depends on how things work under the hood).  In the context of a server with a bunch of people with huge bases that extra memory load could be an issue.  Probably not a big issue right now as the tables are so very short but then look at how the old seed->output table grew when trying to add support for more mods.

A global (shared between all trays) would work better.  Something in init vaguely like:
`if not globalTableCache then
    --read the table from config
end`

Actually something like that but storing all configuration data could be employed so that we only have 1 init check to load the data.  I'm just cautious about how globals work in this context.  More reading and testing (for either case).

RE sanity checks:  It was less my concern about the ability of the FU modders and more if someone does a patch file with horrible values.  But I'm fine assuming expert ability for most values.

Not sure I completely follow note 3...  Did you ask a question about my logic using your code and logic (with a name that you told me not to use in note 1)?

(guessing from previous context)
I think it might be a really old sanity check.  In general some of the sanity checks that are in there are older and from when I was testing things and probing what Lua (and Starbound) could do.

(In general)
Some sanity checks are cheap, either rarely run, or relative to what is lost if something bad gets through.  I find some of them make the code easier to read.  I'm loath to optimise the code so much that only I or an actual expert Lua coder can follow without spending a ton of time trying to follow what is expected.

ProfoundDarkness:
... Colour me surprised... Lua apparently defines division by 0.  Who cares if an incorrect configuration eats all your water items, it doesn't break.  I am so used to languages that don't define it.

What I find funny is that Starbound can't handle that very well in some cases.  For example If you store an infinity into a character and dump said character to json it won't import that exact same json file back in.

ProfoundDarkness:
I'm still not happy with init after line 52, but then I didn't really spent a lot of time working on it today.

Since we are changing some things that were present from before I started messing with the code I have some questions.  What is the point of this?
https://github.com/ProfoundDarkness/FrackinUniverse/blob/branch-Trays_use_seed_data/objects/isn_unifiedgrowingtray.lua#L102-L108

It looks like code that someone put in there a long time ago because you were 'supposed' to but the purpose seems to have passed away with "scriptDelta" in the object file.  Is it because sometimes Starbound can trigger an update too soon?  Is it basically there in case someone mis-configures the object?  I've added some debug tests in before those lines and I cannot see the point of that segment, not one bit.  But then I was surprised when Starbound would **sometimes** store numeric keys as strings (storage variable) and then much crashing would ensue.

I'd kind of like to remove it or if it's there because sometimes Starbound misbehaves I've got a modification in mind.

The other thing is storage.activeConsumption.  I'm not sure what purpose it has and with as much as update() has changed since I started working on this file I don't know if I'm even using it right.

Thanks for helping me to get this thing ready.

whismerhill:
1. `attempt to compare string with number` ok so you found something which doesn't translate from string to number automagically. still remains the fact that : 
`"isn_baseGrowthPerSecond" : "1",` is incorrect and shouldn't even exist in the first place because it's our file of our project. Theorically you might be able to do this : 
`if self.baseGPS + 0 <= 0 then self.baseGPS = 4 end` this should autoconvert to number if required
OR
`self.baseGPS = config.getParameter("isn_baseGrowthPerSecond") + 0 or 4`
though again let me stress out that it should not be required to do so in the first place since we have full control of our JSON.
2. about the power thing, I didn't design the power system, so I'll let others.

ProfoundDarkness:
A little confused... asked about 2 things, got one answer... intuiting that both are related to power and roger on the don't touch.

ProfoundDarkness:
In keeping with the assumption the code I asked about was related to item transfer and/or power I looked at the old code and copied what I saw (for the top of update).  For the other line, I moved it to after 'water' consumption which is where it was in the other code.  The new code places growth in a different place so it's not a terribly similar in logic.

Curious about thoughts on init change (mostly the tail end).

ProfoundDarkness:
... Yeah I said I wouldn't touch but the changes to the logic I made in update() made it tough to tell if I had the activeConsumption line in an equivalent spot in the new logic as it was in the old.  I read through a few bits of code from other devices and re-arranged/added code based on what I found there.

For one thing, the hydroponics tray will turn the lights off when it's not working on something (not just if it has insufficient power).  Basically lights off and you are expecting it to work on something then something is wrong.

The other is I placed the activeConsumption line about where I think it should go given other powered devices tend to have that set to true before consuming power.  It's not conditional on power though, left out in the open like the old code had it.

whismerhill:
the "old" hydroponic tray would work without power just at reduced speed, it was a design choice. No idea where we stand on that matter, I don't think any change was planned.

ProfoundDarkness:
It still works without power, just turns the lights out.

ProfoundDarkness:
By the by, could you let me know if this reaches an acceptable state on paper so I can test it out more.  I don't want to spend a ton (more) of time testing if there are further changes.  I've been play testing but that currently excludes hydroponics trays.

Aside from better power handling for fast forward updates I'm not sure if there is anything else to add or remove.  And the better power handling will have to wait until I understand power better, not just what I read from code but that too.

I mean this is scrapping some but since we have access to the seed data I could make plants use more 'water' if they are wider than 1 slot (for example cotton).

Actually now that I think on it I do have one more commit to do in order to clean up the object files and put some balance notes and observations up on the wiki.  That doesn't change the code though, more to help whoever works on the objects after me.

ProfoundDarkness:
The trays should now behave very similarly to the older versions as far as resource usage and growth time. There are bound to be some variances because the 'old' trays ignore seed data while the 'new' trays use that data.

Though all the growth boost options on liquid items are 0 the code is still there to support a growth boost from liquid if desired.  Basically this makes the new trays act a lot like the old ones but have much greater latitude in tweaking and balancing without having to mess with code.

Old information and how I derived values are on my fork's wiki.

ProfoundDarkness:
I happened on a question when I was playing with the idea of automatically starting the plant over if the player (like myself) put in their seed before their fertilizer.

Is it apt that seedslot, waterslot, and fertslot are going to storage when they are clearly being treated more like self items?  To put it another way I think that line 27 through to 29 should be changed from:
`	storage.seedslot = 1
	storage.waterslot = 2
	storage.fertslot = 3
`
to:
`	self.seedslot = 1
	self.waterslot = 2
	self.fertslot = 3
`

whismerhill:
lucca wanted to take a look AFAIK, but he's kind of busy. From the little we spoke about it I think he thought some stuff was nice, although he had a few gripes with how stuff was coded without using Lua specific techniques (but that was several revisions ago, so might be already fixed).
As for me, I will review it again, probably this weekend.

My opinion is that this thing will get integrated eventually, Sayter's not one of those managers who'll let you work on stuff and then say : "ho in the end I'm not interested". I mean it could happen but it's not his habit, he generally speaks his mind right away so what I mean is that : **if we were not interested, you would know it.**

one thing though, avoid touching any other lua file (such as the power thingy)
at least not now, & not in the scope of this project, 
(sidenote: the current power system is also very resilient & very powerful)

whismerhill:
for storage or self, it doesn't really matter

whismerhill:
ok so I'll say this now, don't work on this more. Lucca's actually doing an entire rework based on your work AFAIK

sayterdarkwynd:
fantastic work guys! this is shaping up to be quite nice.

PR #1224, opened 2018-01-19
FU BYOS custom race support

levithepants:
Added BYOS support for all FR supported races, as well as Draconis.
I didn't test them all, but every one I tried worked.


Hubnester:
Hi, i like the idea of this (for some reason i didn't think of doing it like this). Though i'll probs rearrange the the order of them to match what it is in the modded race support patch (alphabetacally (to make it look better) and i might remove Draconis for consistency and so that people don't say stuff like "Since draconis is in it, why can't you add others?" (i don't want non-FR supported races due to potential issues with it taking too long to load your character)

PR #1225, opened 2018-01-19
BYOS Fixes

Hubnester:
Fixes the issue with the decorative FTL drive placement images. Also makes it so that the zero-g and no oxygen should work on other peoples BYOS ships. Re-does the universe_server.config.patch to make it match the format of the patch mod + removes non-FR supported races (NOTE: AVAILI WILL NOT BE ABLE TO GET BYOS SUPPORT INCLUDED IN FU DIRECTLY (NEEDS A PATCH MOD))


Hubnester:
Also i was a bit lazy with the universe_server.config changes and will probs improve it by removing races that can't be supported in FU (i didn't feel like doing much coding today (at least for now))

PR #1249, opened 2018-01-30
Allow more flexibility when adding collections

ColonolNutty:
This modification allows more flexibility when it comes to adding custom collections.

It takes the data property from the button and uses that to find the collection list, instead of hardcoding "customCollections" and "customCollectionsVisible"

As long as "customCollections" is at the start of the name, it will be considered a custom collection.

For example:
{
//button
"data" : "customCollectionsSomething"
}

"customCollectionsSomething" : [ "someCollection" ]
"customCollectionsSomethingTitle" : "Some Collection"

### DIFF
--- a/interface/scripted/collections/collectionsgui.lua
+++ b/interface/scripted/collections/collectionsgui.lua
@@ -44,19 +44,23 @@ end
 
 function populateList(collectionName)
 -- BEGIN CUSTOM CODE
+  local isCustomCollection = collectionName ~= nil and string.match(collectionName, "customCollections");
   local collectionName = collectionName or widget.getSelectedData("collectionTabs")
-  if collectionName == "customCollectionsVisible" then return end -- special case: do nothing
+  if collectionName ~= nil and string.match(collectionName, "Visible") then
+    return
+  end -- special case: do nothing
 
   widget.clearListItems(self.customList)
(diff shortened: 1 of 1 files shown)


sayterdarkwynd:
comment from one of my devs:

"despite saying "more flexibility" it also seems to enforce that custom collections have customCollections in their collection names

additionally, if any collection has the word Visible in it, bam exit

I'm not a fan of code behavior being based off of arbitrary input"



its not a big deal, but a valid input on it. Thoughts? Also, is it not possible to simply do a patch to add  "your_collection" to the existing "customCollections" : []  list in the collectionsgui.config file?

ColonolNutty:
That will inevitably lead to extreme lag when loading the collections

sayterdarkwynd:
fair point. collections fucking suck for optimization. nothing else loads as slow, i dont think. Which, frankly, baffles.

ColonolNutty:
I'm thinking of overhauling the collections window so the "buttons" act similar to how they do in the Recipe Book in MFM, where it can be more dynamic and the positions won't matter as much

ColonolNutty:
Yeah, this works with FU how you have it now. I tested it in game to verify.

sayterdarkwynd:
Yea, collections could use more love.

PR #1256, opened 2018-02-02
Change customCollections to customCollectionsFU

ColonolNutty:
This is to make it easier for mods to check whether or not FU has modified/added to the collectionsgui.config.

Instead of having to check whether the FU collections button has been added via the image it uses, a patch file can check for the customCollectionsFU to have been added.

Since FU changes the size of the collectionsgui.config window and the button locations, this will make it easier to provide compatibility for FU without messing up the UI Frackin' Universe uses, but also allow mods to support the old collections window size/positions for those users not using Frackin' Universe.

(It is possible to do without these changes, these changes simply make it easier, although I haven't accounted for any of the FU addons)

### DIFF
--- a/interface/scripted/collections/collectionsgui.config
+++ b/interface/scripted/collections/collectionsgui.config
@@ -164,12 +164,12 @@
           "baseImage" : "/interface/scripted/collections/tab_fu_deselect.png",
           "baseImageChecked" : "/interface/scripted/collections/tab_fu_select.png",
           "pressedOffset" : [0, 0],
-          "data" : "customCollections"
+          "data" : "customCollectionsFU"
         },          
         {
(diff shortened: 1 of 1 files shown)


sayterdarkwynd:
people are complaining that they can no longer open FU collections since your change. I havent checked and will do so today.

ColonolNutty:
Have they given reasons why?

ColonolNutty:
Like I said above in the description anyways, the changes in this pull request may cause issues with other Frackin' Universe addons. If so, then sure, go ahead and revert these changes only. DO NOT REVERT THE LUA CHANGES AGAIN without providing some proof such as logs, errors, or stack traces, it will be irresponsible to do so.

PR #1343, opened 2018-03-28
Update byos to support variations of default ships

ColonolNutty:
This way other mods can modify what the default ships capabilities are without having to modify or extend the fu_byosai.lua script.


ColonolNutty:
An example of a modification: 
```javascript
{
  "shipLevel" : 1,
  "capabilities" : [ "teleport" ]
}
```

ColonolNutty:
Tested with just FU installed

PR #1358, opened 2018-04-14
Desc: Little Good Berry grants 10% damage not 15%

XDGrangerDX:
There seems to be a problem with

```
"effects" : [ [
      {
        "effect" : "damagebonus",
        "amount" : 0.15,
        "duration" : 120
      }
```
not applying, someone else investigate?


Kherae:
that "amount" variable is ignored. that's why. we need to use damagebonus2.

sayterdarkwynd:
already fixed this locally :)

PR #1385, opened 2018-05-11
fix typos and clarify org soup interaction

locketeerian:
"Strait" is a typo.
"Curved" sounds more natural than "Bend" in the middle of a sentence.
The org soup interaction is more descriptive using "hunger" instead of "food". "Restoring food" can be misunderstood as "this soup is undoing food spoilage", which it does not (according to my understanding). It only replenishes the player's hunger meter.


sayterdarkwynd:
looks good

PR #1401, opened 2018-05-26
Exclusion List Edits

Xilosc:
Removal of alloys from the exclusion list as they can't be processed in the Blast Furnace or Arc Smelter so wouldn't be redundant and would be able to be processed here

### DIFF
--- a/objects/power/manufacturing/mfgstation.lua
+++ b/objects/power/manufacturing/mfgstation.lua
@@ -36,13 +36,8 @@ local exclusionList = {
   pyreitebar=true,
   isogenbar=true,
   xithricitecrystal=true,
-  carbonplate=true,
   fuamberchunk=true,
-  advancealloy=true,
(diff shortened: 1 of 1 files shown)


sayterdarkwynd:
I don't *want* alloys to be processed at this station, at all.

Xilosc:
Alright, they currently have no way to automate but honestly there's not much point in automating such rare resources anyway. The only one that you would use lots of would be carbon plates. I know automation is more of an extra anyways. I had just added everything from the first tab of the furnace to exclusions, then later realized you couldn't process alloys anywhere. I would still like to make carbon plates processed here since you can end up making lots of them for things, the others not as much of a concern since they're mostly used for weapons or armor.

sayterdarkwynd:
I suppose it breaks nothing by allowing it, since they still require the resources. Fuck it. Why not.

PR #1420, opened 2018-06-09
Adds patches for all the techstations on the Steam workshop as of 2018/06/07

xn-5o8h:
Ran some scripts to parse the whole workshop and make patches for all the missing techstation objects for S.A.I.L.'s interface. Should have every one of them unless I messed something up.

Since all of those patches are automatically created, and that probably most of the mods who add them are not-really-used species (meme mods etc.), this might not be worth adding. Still, I doubt they'd cause any problems, and it won't hurt to have more compatibility.

(Also, because FU has a default priority, Starbound doesn't have any reason to load it after the mods who add those techstations, so patches might not be applied consistently.)

### DIFF
--- a/object/ship/avianapextechstation/avianapextechstation.object.patch
+++ b/object/ship/avianapextechstation/avianapextechstation.object.patch
@@ -1,0 +1,26 @@
+[
+    [
+    	{
+    		"op": "replace",
+    		"path": "/interactAction",
+    		"value": "ScriptPane"
+    	}
+	],
+	[
+		{ "op": "test", "path": "/interactData" },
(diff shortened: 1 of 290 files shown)


sayterdarkwynd:
wow. that is not a small amount of bullshit ! Thanks man. I dreaded doing this lol.

Still, we should pack all of the stations in an addon that Includes all the race mods, and FU, so that we can make sure the SAIL works. but meh.

xn-5o8h:
Would this fit the bill? https://github.com/xn-5o8h/FUSAILModdedRaceSupport

I can easily keep it maintained. It's only existing in that repo for now because I don't know if you want to have something specific going on with it and FU or whatever, waiting for your feedback.

I could also make a PR to remove all the patches from FU, since having both seems unnecessary.

PR #1422, opened 2018-06-09
Bees candles now wired

Alepod:
So you could turn it on/off


sayterdarkwynd:
You have loaded, tested and verified these work okay?

Alepod:
Yes, it works properly

PR #1441, opened 2018-06-30
Not-so-micro treehouse dungeon

aelfgifu:
Changes treehouse5 from a microdungeon to a regular dungeon, giving it a better chance to spawn. (Aelfgifu)
Adds a barn and hidden room to the treehouse (Skies912)


aelfgifu:
(p.s. If you go to diff settings and check ignore whitespace changes, it's not really that much changed in terrestrial_worlds.config.patch. That freaked me out for a minute. Thx, Atom.)

PR #1448, opened 2018-07-09
bgm pool changes

HALLEYSHURA:
-changes to the bgm pools of many FU biomes. should be more variation between bgm for new planets now


sayterdarkwynd:
cool. would have preferred this for FM though

PR #1470, opened 2018-07-27
shiny

Kherae:
first pass at biome weather messages. affects weathers not in the /new/ directory, that have radio messages. They will no longer spam you, as the cooldown is now a resource on the player. Instead, the cooldown (60s) starts after you lose the weather effect. weather in /new/ will take further work, as the mess is further ingrained.

removed the isn_sharedobjectscripts.lua, moved some functions to effectUtil.lua. 
Note: atmospheric regulator and filter need a warning stating that not all effects will work. This is because you cannot send stat-groups via message handler, only effect id strings.


Kherae:
also refined the marvin effect a little, nothing big.

PR #1485, opened 2018-08-04
Removed placeholder research tree

zimberzimber:
As currently is serves no purpose, and it will just hang in there and confuse people


sayterdarkwynd:
would honestly rather just leave it there so we dont need to re-add this stuff when the time eventually comes.

PR #1489, opened 2018-08-05
Tech Descriptions

Kherae:
Changed Power Jump fall damage reduction to bring it to sane levels. Similar in Flutter Jump techs and Gravity Jump, which reduced fall damage by 99%. Power Jump reduces fall damage by 20%, Flutter Jump techs by 30%, 40%, and 50%. Gravity Jump by 75%. Note that these jumps still allow you to jump large distances safely due to their gravity reductions...and wall slide.

I also changed many tech descriptions to make them more concise and informative, make descriptions fit in the description bar on the tech selection UI, and display their relation to preceding techs. I also changed displayed Hotkeys to be more reliably color coded: Green activates a main part of the tech (see physics field for why I say -a- and not -the-), yellow activates a utility ability that requires what it follows to be active (again, physics field). Red activates an offensive ability, such as sphere bombs.


Kherae:
added a fix to /objects/vanity/cute/cutevase/ objects that had trailing commas in colonytags, causing them to fail to load. 
also added a fix to bridges, so they should properly clear their object lists

PR #1494, opened 2018-08-07
useful

Kherae:
Set default selected categories for several ui panes, to lessen the freezing when people load stations with a high number of unlocked categories.


Kherae:
added: fixed 5 throwable items missing categories. changed Sifters tutorial quest completion text to be less awkward.

PR #1523, opened 2018-08-20
precursorshotsmall2 fix

Shinigami-Apples:
requested by Kherae


Kherae:
(for record) fix for:
[Error] Could not load image asset '/projectiles/guns/precursorshot/precursorshotsmall2.png:5', using placeholder default.
(ImageException) call to subImage with pos (45, 0) size (9, 5) out of image bounds (53, 5)
 sadly I cant fix this, lack of an editor. file just needs to be extended 1 tile to the right

sayterdarkwynd:
already fixed that

PR #1528, opened 2018-08-25
nerf hammer

Kherae:
rather that nobody hears of this one til it's too late. heh. nerfing ITD and grabber recipes to require teleporter cores, so they're more costly to make.


Kherae:
adding:
fixed duplicator animation file typo
changed duplicator slots to be ordered 1 2 3 instead of 2 1 3. people will need to move their items.
duplicator will now attempt to stack outputs into the item input slot if the output is full, before simply dropping them.

PR #1556, opened 2018-09-24
Quest Changes

Tamamo89:
-Starbooze Quest now followup to the next quest.
-Black Glass quest deprecating arcsmelter quest now unlocked by blast furnance quest.


sayterdarkwynd:
the booze quests are intentionally not following up, so that people use the Tricorder.

Tamamo89:
do you want me to do the same with the bee quest then? they're in the same category.

sayterdarkwynd:
i dont want *any* of them doing auto-followup. the bee ones are probably fine as-is though.

Tamamo89:
@sayterdarkwynd 
so you want the bee ones to be left as is with auto-followup.
Kay. I'll revert the changes to the booze quest then. Then will this be okay to merge?

PR #1605, opened 2018-11-11
FU weather fixes

Wannas16:
This is a somewhat extensive revamp and clean-up of the various FU weather .lua files. I've tried to make their behaviours a bit more consistent in preparation for a common library implementation. While doing the rework, I believe the following bugs/issues have been fixed:
* All relevant immunity stats should now protect against weather effects, even if they are not the selected blockingStat (previously, only the blockingStat or resistance threshold would work).
* Some of the issues with spamming the SAIL warnings should be fixed (for example Aether warnings when beaming back to ship).
* Weird behaviours where the weather or radio-message timers ticked down multiple times have been removed.
* Movement penalties when at low health in Desert (25% HP) and Jungle (33% HP) biome weathers are now correctly implemented.
* Fixed randomised messages for Insanity and Pressure weather effects.
* Internally, effects should hopefully no longer be expired and re-applied every tick.

The system in general would greatly benefit from the use of a common library file, since the .lua files are 70-80% repeated code. So consider this a short-term fix. I've tested each of the weather effects by beaming down to planets and testing debuffs, immunity stats and resistances. But I would of course recommend that you test it yourself before merging.

### DIFF
--- a/stats/effects/fu_weathereffects/new/aether/aetherweathernew.lua
+++ b/stats/effects/fu_weathereffects/new/aether/aetherweathernew.lua
@@ -193,25 +193,22 @@ function update(dt)
 	end
 
 	if isEntityAffected() then
-		self.biomeTimer = self.biomeTimer - dt
-		if (self.biomeTimer <= 0) then
-			-- Aether effects are worse at night (while above ground).
-			if not (daytime or underground) then
-				setNightPenalty()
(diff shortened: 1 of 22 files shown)


sayterdarkwynd:
looks good. will start testing it after tomorrows release so nothing gets missed before a push :)

Wannas16:
Updated with this week's release.

PR #1607, opened 2018-11-13
Crystal Fire is a Good Substitute for Pepper

Llarys:
Fixed the fire shard gun to no longer use the appearance of the ice shard gun.


sayterdarkwynd:
already fxied on my end. thanks though!

Issue #1620, opened 2018-11-29
ArCon Armor dyed colors appear to be wrong


AugiteSoul:
![image](https://user-images.githubusercontent.com/16467639/49243304-4b4eae80-f40d-11e8-9d28-de8de5f497de.png)

Dyeing the whole set using the vanilla black dye make it appear green instead. Is it intended?

PR #1622, opened 2018-12-02
 Fix for #1620, dyeable blister armor, fix previous pr

loudwhitenoise:
- Fixes Dye on ArCon Armor, will be green by default
- Fixes Dye on BattlePacks, comments within file indicate colour decisions
- Dyeable Blister Armor
- Dyeable Blister II Armor


sayterdarkwynd:
you do realize you could make these fixes FAR easier right?

open armor file in program. re-paint it to vanilla default colors. Done.

Then you don't need to hunt for Hex values.

in any case, im not going to commit these till youve tested them. Also, i am 100% not making ArCon green by default, period.

sayterdarkwynd:
also, the arcon armor is *already* dyeable....not sure why it was changed at all.

sayterdarkwynd:
did you correct the issues and test them out, loud?

PR #1629, opened 2018-12-09
Nightar Ship Update

AbsoluteXeroEX:
This primarily updates the Nightar ship structures with a new blockkey configuration, as well as adding the 'lit' layers (so the details of the hull and light trims can be seen *properly*). I plan on revamping the Nightar ship structure and graphics entirely in a future update to the core Nightar mod later on (because the Glitch-style layout can eat a big ol' bag of dongs), but I at least wanted to get these updates in.

One small change was made to the Nightar starting treasure pool regarding the main/broken weapon (changes **Broken Desinium Sword** to **Broken Nightar Saber**, supplemented with appropriate 'repaired' version and corresponding recipe); if you'd rather keep that the way it was, I'll be happy to revert that change and re-submit.

(Quick edit: I realized after the fact that I had removed the files for the Broken Desinium Sword, added those back in to avoid potential conflicts).


AbsoluteXeroEX:
Added missing 'brokennightarfuelhatch' object; should be good to go as of this commit.

sayterdarkwynd:
only thign i see issue with is the sabre. Not sure why that was changed. The existing one in FU was fine as-is. If there's a reason, cool, explain and I'm sure it'll be fine :)

sayterdarkwynd:
I see. I omited the nightfort entirely on purpose because it , well, had little to no actual purpose for inclusion (in its original Nightar mod format) and was kinda lame. If that has been addressed, I may revisit it. 

If it is to be included in FU, I think that Lore should suit it first, otherwise its just there for the hell of it. In the current mod youve got working, how/what/why is the nightfort incorporated? Does it have any purpose for existing being "cool place to visit with a couple of rather drab quests that serve no large-scale story purpose" ?

I've no objection to it being a more interesting thing to get overall. However, I've altered the lore focus of the Nightar to one-handed blades rather than broadswords, which is why i have the two short blades as starting weapons in *addition* to a broadsword now. giving the player options on how they use swords seemed the best case scenario, as this way people can use whatever they wish!

AbsoluteXeroEX:
Ah, gotcha. From a story standpoint, it was based on a more *vanilla*
aspect, which was a temporary thing until I solidified the lore a bit
better; basically it was to eliminate the threat of the Night King's new
conquest, which is a plot that admittedly I'm still fleshing out proper. I
can go ahead and revert the weapon change until I get the mission bits
smoothed out from an FU perspective, which will likely include a complete
redesign.

sayterdarkwynd:
"Night King" sounds lame as fuck (original lore is all awful, really)

. I think a more viable social/political structure needs to be put in place. I'm leaning heavily on a matriarchal society, based on a commune of individuals from various groups/clans whatever, rather than a singular leader. 

Logically speaking, there is no fucking way a *single person* would be in charge of an entire civilization.

Instead, given that they would have , by now, spread to numerous worlds, youd have a much wider variety of cultural differences among their kind. This would mean that each world would need to have its own governing body. There might be an underlying "core" government...but I doubt very much any true spacefaring civilization would be viable with despotic or monarchistic rule.

AbsoluteXeroEX:
That was the point I was trying to address with the re-write. Some guy
*called* himself the "night king" and basically wanted to start with all
Nightars, and eventually everything else. It's been something I've been
trying to properly work out for a while, though I've been a bit hindered
with other projects that are taking a higher priority. I'll revert the
weapon change in a bit, though, once I get to work.

AbsoluteXeroEX:
Nightar starting pool has been reverted to its original state; should be good to go now.

AbsoluteXeroEX:
Understandable mate; I've been at work for 30 minutes now, and I've been getting slammed on what would otherwise be a boring-as-shite night.

PR #1648, opened 2018-12-23
zbSail Update

rylasasin:
Add United Systems: Expanded and Extended Story missions.


sayterdarkwynd:
but...they arent fully compatible with FU and shouldnt be used along with it in the first place....

but okay...

PR #1654, opened 2019-01-01
Fix baseDamage scaling

laz2727:
Currently, gunfire doesn't scale baseDamage at all. Which is a problem because some vanilla things, like grenade launcher alt, use it by default.


sayterdarkwynd:
all you did was add brackets. That won't change anything.

self.baseDamage or (self.baseDps * self.fireTime)

is the same as

(self.baseDamage or (self.baseDps * self.fireTime))

not sure how this does anything?

laz2727:
Due to how lua logical operators work, the expression never calculates, because baseDamage exists and or returns that.

PR #1676, opened 2019-01-17
Precursor Spawner Change

Hubnester:
- Fixes an exploit with the precursor spawner


Kherae:
Khe disapproves. renders spawner useless. let's make it require the neutronium stuff to even function, and up the fuel cost, instead. People can prove they are worthy of this power, that way.

Hubnester:
The main problem with the spawner currently is that you can automate precursor loot and possibly elder loot, and a problem with your idea is that upping the fuel cost makes most other monster loot not worth the effort and cost

PR #1764, opened 2019-03-19
Fix lag with tricorder

XanTheDragon:
Changes done to achieve this:
1) Increase scriptDelta to 10 (from 1)
2) Remove UI data population from `update()` since the data is static and this only wastes performance


XanTheDragon:
Doing edits to this PR. Will resubmit.

PR #1872, opened 2019-07-04
Fix for Chinook body being set too far back

lamialily:
This is rather subjective but I figure I'd put it forth anyway. The Chinook mech body is set so far back that the entirety of the far shoulder is visibly detached from the main body, so I put it forward four pixels to accomodate.

And while I'm on the subject, is there anywhere development is discussed so I can check in before taking a shot in the dark like this?


sayterdarkwynd:
there is, but youd need invitation to the dev channel, the master document or the hack n plan, all are reserved for official staff

lamialily:
Fair enough. Maybe I'll just focus my efforts on new content then.

PR #1901, opened 2019-07-26
Go away fools

Davee47:
Changed the "Unused" research tree to be bright(?) red and the text to Unused, Do not use! so people stop using it. 
Had someone complain about being unable to read the word unused...


sayterdarkwynd:
heh it was turned black specifically to be unreadable and the header text within says DO NOT USE.

people are stupid.

PR #1905, opened 2019-07-29
grand sovereign temple pgi fix

Kherae:
Fixed a PGI in a certain bookcase in the Grand Sovereign Temple


Kherae:
added: fix for fenerox species patch using indexes, resulting in undesirable behavior depending on mods.

PR #1906, opened 2019-07-31
fringe case fix

Kherae:
tentative fix for a fringe case error with the poptop cavern door which occurred when a player somehow didn't have a unique ID. Instead of erroring, it will just print a warning to the log. Also commented out an unused scan from the update block and its variables from init, and prettied up indentation.


Kherae:
added:
fixed poptop forkbomb

PR #1916, opened 2019-08-10
lots

Kherae:
lost and found merchant is now properly a veluu. Mew.
shifted aether augment unlocks off items the recipe doesn't use (effigium, tritium), onto aether liquid
added missing tags to some pyreite weapons
hydro/growing trays will only consume seeds, fertilizers and liquids from their respective slots. actual impact will be minimal and largely only noticed on double soup setups
mysterious balls renamed.


Kherae:
added:
removed poison/rad icons from harpoon guns

Kherae:
added:
added 'hunting' info to harpoon guns

Kherae:
added:
fixed crankgun type icon

PR #1925, opened 2019-08-19
big list

Kherae:
water walk and item magnet fix for multiplayer claimed worlds.
some small fixes for effectUtil
reverted 'fix' to transferconfig which actually broke it. oops.
changed an artifact's category to be artifact instead of crafting material
added special status immunity to some boss monsters, including big fat spoiler boss
added deathbomb effects: percent max health to various currencies (with support for items), 'hydra' effects, npc 'corpse looting' (non-npc items only)
trolol, vulnerability, and cultistshield now also affect certain elithian races damage types.
added a monster inspector item that will display a monster's description, kinda like the scan tool, recipe unlocked by picking up iron.

below this line, do not include in change notes!

changed killpod's effect in modified pillar and warped regulator from fu_death to killpod.
warped regulator recipe unlocks from pillar, until a workaround is managed. 

added a ton of effect options, including many of the above listed effects, as well as fine tuned some. (autowarp in particular. it now functions reliably. does still need vfx)
one example: sticking phase matter in results in enemies being invulnerable and unable to do damage, for the most part.
sticking certain voxel items in (highest vanilla/fu ones) allows you to kill enemies to make use of those percent max hp to currency effects. madness and essence are at 1%, so not terribly exploitable.
hydra effect involves shoving a precursor spawner in, and spawns a copy of the enemy every time it dies...unless it has any spawning actions in its behavior config. because hemogoblins.


Kherae:
the deathbomb currency effects would make nice weapon effects, if that's possible. like pirates looting pixels or something.

PR #1941, opened 2019-08-29
Add more item EE for Starbound support

kju8:
want to manage all metals on the EE table


sayterdarkwynd:
absolutely not allowing end-tier metals in that. no sir.

kju8:
okey, i'd like to make end-tier support with personal patches.

sayterdarkwynd:
I understand its "fairly" well balanced , but Noxcium is absoultely not meant to be common or easy to come by, and should only be procured via nocxium or manually crafting. The others are less problematic. Aether is a no because it isn't an ore, its an alloy composed of several metal types.

the rest are less of an issue, as they can be readily mined.

PR #1986, opened 2019-10-03
Item Tooltips pt. 2

Hullaballoonatic:
* completed expansion of item tooltips for all generics and liquids
* added chemical formulae to all real world chemicals
* updated terminalGui.lua search filter to ignore the color codes in item names when comparing to filter text
* removed changes to some fruits due to issue with FR compatibility
* some fixes and improvements to tooltips
* some files were prettified
let me know if you find any issues. i searched a lot already.


Hullaballoonatic:
Just noticed a change I needed to make.

PR #1994, opened 2019-10-06
up

Kherae:
Removed nonexistent greenguardcodex8-codex entry from largedungeons.treasurepool.patch
Updated network guides.


Kherae:
added: removed racial descriptions from atmospheric condensers, which caused scan functionality to not work for certain races.

PR #2006, opened 2019-10-14
stuff

Kherae:
Fixed an oops in madnessdata, due to khe using method style incorrectly.
nerfed energizing augment. no longer regenerates except when energy is naturally regenerating.
tweaked physics field. fall damage protection is now a toggle, and no longer negates movement.


Kherae:
added: Fixed gene design lab not unlocking previous stages' unlocks.

PR #2008, opened 2019-10-18
memento mori fix

Kherae:
fixed fringe error that occurs when you swap to the crystal skull while firing, caused by derpy variable handling.


Kherae:
added:
fixed missing audio error in 9 biomes due to a missing underscore
hopefully fixed the embalming table printing a rather harmless error about a missing frame
corrected a case difference resulting in a madness radio message not playing

PR #2011, opened 2019-10-23
Fixed mispelling of tungstenore

DragonWolfLeo:
Just fixing a little bug I encountered


sayterdarkwynd:
i actually fixed that already but thanks:)

PR #2032, opened 2019-11-05
Text flow - chemistry & metals

aDrunkLittleDerp:
• Changed some tooltips to prevent text from escaping
• Added some punctuation and capitalization


sayterdarkwynd:
not sure how i feel about arbitrarily changing their tooltip to that of fossil, but it shouldnt affect much.

PR #2033, opened 2019-11-09
Name reversion

Argenttip:
Reverted name back to Lovelius Smythe from Occult Researcher


sayterdarkwynd:
No. I changed it for a reason: There would not be 900 people named Lovelius Smythe running around the universe with the same looks and profession.

PR #2055, opened 2019-11-23
Orphan paste rebalance and recipe addition

SentientCookie:


SentientCookie:
raised price on jarred fetus;
rebalanced orphan paste and added madness to its consumption;
added a recipe to craft 5 orphan paste in the kitchen by using one jarred fetus.

PR #2068, opened 2019-12-08
Bolbohn and Duneseed Fix didn't take previously

Argenttip:


sayterdarkwynd:
But now the loot is doing:

guaranteed seed and produce, and a 45% chance for an *additional* seed. That isn't correct.

Argenttip:
Shouldn't that be the case, guaranteed seed and produce drop with a chance of an additional seed? 

Granted, the seed drop could be lowered from 45%.

PR #2081, opened 2020-01-25
Shoggoth head color fix

Ickura:
Someone messed up the colors bad


sayterdarkwynd:
or they are exactly the colors they were designed to be.

Ickura:
Of course, there was a duplicate hex value for the shoggoth armor set and now they look ugly and inconsistent to the dyed versions. Thats should clear things up.

PR #2086, opened 2020-02-05
Update quietuscaptain.activeitem

TheFirel:
Noticed Captain's Blade was changed to correct it, but resulted in another error being made.
This change unifies it with the likes of the "Assassin's Blade".


sayterdarkwynd:
Actually, the current name is not incorrect at all. It simply changes the intent of the word Captain, implying that this is a blade made for captains, rather than a blade that belongs to one. I changed it to this intentionally so that no confusion remained as to the intent. Removing the possessive modifier clarifies that.

PR #2110, opened 2020-03-02
Tungsten and titanium broadswords

rylasasin:
It has come to my attention that there are no tungsten or titanium level broadswords. 

I have fixed this.

PS: The reason they have "longsword" in their filenames and not "broadsword" is because the tungsten and titanium longswords have "broadsword" in theirs, for some reason. I would have fixed this, but that would have broken those weapons.


sayterdarkwynd:
neither exist because vanilla has their own, so we never bothered due to redundancy.

PR #2114, opened 2020-03-08
Apex additions to autopsy table and disabling printing of Apex life support pods.

Argenttip:


sayterdarkwynd:
I misunderstood.

The agaran pods would be fine.

Using a scalpel on technology without a corpse in it doesn't make any degree of sense.

Argenttip:
The regular Apex life support pod has an Apex inside. I can change it so putting it in the autopsy table produces an empty life support pod and a corpse.

As for the brains in jars, I figured this would simply represent opening the jars.

Argenttip:
Agaran pods have been changed to non-printable to remove exploit of endless fungal ichor.

PR #2115, opened 2020-03-08
fix "unknown treasure pool" for inspired paintings

adaxe:
I think those entries should be "item" instead of "pool". 
Currently it create "unknown treasure pool" error, as shown in the log:
https://pastebin.com/gHA4NzXM


sayterdarkwynd:
yep! thanks.

PR #2118, opened 2020-03-16
Frackin Vanilla Dungeon Difficulty Update

rylasasin:
**Vanilla Dungeon Difficulty Update: Erchius Horror, and The Great Pagoda**
Both of these vanilla dungeons have been upgraded in difficulty, in two completely different ways.

**Erchius Horror:**
In vanilla, this was brought down in difficulty from its Giraffe incarnation simply because it went from optional mission to a mission you _needed_ to do to get your FTL drive fixed in 1.0.  Frackin however, reverts it to optional status (Dark Cavern now takes the story-essential role EH used to have.) To reflect this, I have restored this dungeon back to its Giraffe level of difficulty to encourage players to do a bit more exploring before they jump into it, and to give players a taste of what to expect in terms of difficulty from Frackin content.

You will now require decent armor and weapons to complete it.

- Mission is Tier 3 now, instead of 2. It's the same level of difficulty as Sunset Riders, but the Moontants don't cheese you with sniper rifles.
- Moontants have marginally more health, and do more damage. They also have FU resistances and vulnerabilities.
- Mutant Miners are much more tanky and more deadly than before. Early level players WILL need the miner's help in taking them down. They are much more "miniboss/microboss" than they were before.
- Surviving non-mutated miners have SEVERELY reduced health. They can actually die now if you're not fast enough to save them.
- Erchius Horror does more damage and has more health (side effect of tier change). You'll only notice the latter however if you try to take it down the hard way. The mining laser still take 3 hits to take it down (its damage output has been increased to compensate.)
- Overall, this mission is more like it was back in the Giraffe days.

**The Great Pagoda Library**
The Great Pagoda Library always had one problem: It was _nothing._ It was a boring mission with very little combat and no puzzle solving. Not a very good way to introduce the story's secondary antagonists.  Basically, it just felt... unfinished.

To fix this, I gave it more stuff.

- Cultists are everywhere now. This mission is no longer a "walking simulator," there is no longer a single room where there isn't something going on.
- Cultists come in progressively different flavors. They use pistols (and sometimes machine pistols) to start with, later you'll encounter ones that use MPs exclusively and have critters to distract you, later you'll encounter them with assault rifles, shotguns, and sniper rifles, and finally at the bottom you'll encounter ones with Rocket/Grenade Launchers.
- Cultists can also be seen with (a limited selection of) Frackin items. Don't worry, this isn't Eden World level of insane, they'll be carrying Zerch stuff *at worst*.
- Cultists can sometimes be seen wielding katanas. When in Japan, and all that.
- Snuffish are now joined by Lilodons and Gosmets.
- Optional miniboss in the flooded room.
- Dolphins.


sayterdarkwynd:
neat. ill gladly test this now that things are calming down, soon.

the erchius changes may not last long though...im making good progress on rebuilding the mission in Tiled.

PR #2121, opened 2020-03-21
Update antimite.recipe

Argenttip:
Add bee flowerheads to the basic antimite frame recipe.


sayterdarkwynd:
already did this, sorry!

PR #2130, opened 2020-03-27
stuff

Kherae:


Kherae:
Madness builder tenant fixed
telebrium sniper muzzle offset adjusted
hops quest no longer erroneously references terramart

Kherae:
added hops recipe. this addresses it only being available from the wild, after the prop pack update in FU a year ago.

PR #2134, opened 2020-03-28
inventory count fixup

Kherae:
goofed on Khe merchant's max items. fixed, hopefully. had a max of 20 when there's 42 potential items in her stock


Kherae:
Added:
Fixes towards paralysis effects causing NPCs to not quite die.
Deathbomb effects no longer function on BEEEEEEEES.

PR #2137, opened 2020-03-30
quest fixes

Kherae:
Changed most of the FU quests that consume items to require the player to turn them in to the quest-giving NPC, to prevent a vanilla bug that resulted in the quest consuming the item without progressing if you had it when picking up the quest.


Kherae:
added:
Fixed apiary rivalry check function
Corrected Mulch not displaying its seed use reduction

PR #2153, opened 2020-04-05
quest fixes

Kherae:
updated several FU quests to point the users where to go, as well as adding questgivers/turnins for those that lacked them.

### DIFF
--- a/interface/objectcrafting/pethealingauto.config
+++ b/interface/objectcrafting/pethealingauto.config
@@ -1,0 +1,50 @@
+{
+  "gui" : {
+    "background" : {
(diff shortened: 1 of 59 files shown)


Kherae:
added (put in solid testing first!): 
ai chips and warped ai chips now unlock automatic pet healing stations that are item network capable. sweet quality of life!
warped ai chip also unlocks a much faster, if lossy, pixel compressor. (I hate vanilla crafting UIs sometimes)
added another funny effect to the warped regulator, one that nobody will see unless they have spent 20 million pixels on one item.

PR #2155, opened 2020-04-06
stuff!

Kherae:
Minor conversion to certain conversion effects, making them more flexible to use
self-cannibalism now actually does something, again. When it hits: players without hunger enabled will take damage over time, while players with hunger will instead sloppily take bites out of their health whenever their belly is even slightly empty...and they will get hungry faster. additionally, players will be immune to healing during the duration.


Kherae:
added: switched veilendrex from nonexistent shadowcloud projectile to shadowelementcloud

Kherae:
added: fixed precursor girder tiles. Katrina helped with images.

kukagres:
There are two other blocks with the same names:

precursorfloor/precursorfloor2 --> "Precursor Floor"
precursorpanel/precursorwall --> "Precursor Panel"

Kherae:
added: renamed second precursor floor and panel matitems to match their material

PR #2160, opened 2020-04-08
stufff

Kherae:
fixed salvage component sorting in ITDs
fixed avali door nodes
updated some avali descriptions
added additional code handling to some extraction stations, to prevent items being voided in some cases.
added handling to various stations that makes their wired outputs turn on and off based on whether they are working or not.


Kherae:
added: tentative fix for a lua conversion error in  flying.lua (vanilla bug)

Kherae:
added:
added condition 'is player' to a few madness spawning effects
added condition to a large range of effects' 'status text' to prevent them from firing that portion on nonplayers, to reduce screen spam in certain cases (madness paintings in DF2, as example)

PR #2164, opened 2020-04-08
 changed some description and shortdescription on blocks and crafting materials

kukagres:


sayterdarkwynd:
while i dont see a need to add "rod" in any of those...fuck it, why not.

PR #2172, opened 2020-04-12
Bucket removal

Argenttip:


kukagres:
I think you are missing the recipe and the reference in "player.config.patch"...

PR #2173, opened 2020-04-12
a fix to armor resistance

adaxe:
Currently the Friendship Mask doesn't affect physical or electric resistance at all.
I think it is save to copy the codes of race bonus in FR, or of some exsisting armors that provides resistance correctly (Geist armor etc.), which looks like

"stat" : "physicalResistance"
"amount": 0.2 

instead of 

"stat" : "physicalResistance"
"baseMultiplier": 1.2

### DIFF
--- a/items/armors/uniques/murdermask/murdermask.head
+++ b/items/armors/uniques/murdermask/murdermask.head
@@ -43,11 +43,11 @@
     },    
     {
       "stat" : "physicalResistance",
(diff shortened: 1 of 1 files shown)


sayterdarkwynd:
No. You simply misunderstand how it is meant to work. It amplifies the *current* amount of your physical resistance by 20%. If you have 0, that means 0.

Still, this is technically fine but not exactly the goal.

PR #2177, opened 2020-04-15
Honey changes

Argenttip:


sayterdarkwynd:
almost. you missed something in the ice honey within the centrifuge table.

namely, rarest (or was it rare) having a 2, rather than a 1. Number must match total entries of that rarity.

Argenttip:
Okay that should be fixed now.

PR #2182, opened 2020-04-20
 Changed recipe for "Wine", description and short desc on some elements

kukagres:


sayterdarkwynd:
Why did you cut the text from liquid aether? Nothing was wrong with it.

kukagres:
More short better, right? And the reference to lava ...... can be deduced from the lab directory.

sayterdarkwynd:
Yea, but that doesn't mean flavor text needs to be stripped. Apart from that the rest is fine.

PR #2187, opened 2020-04-29
Changed species tag in codexs for pharitu

noname2tellu:


sayterdarkwynd:
nope. There is no species file named "pharitu". Juux were there, and were replaced with Pharitu. To avoid anyone crashing, the original species name was kept. Appreciated, but i cant use an ID that isnt utilized anywhere else, as that will cause confusion later

PR #2195, opened 2020-05-04
fix!

Kherae:
fixed cute sconce, presents, drapes


Kherae:
added:
added debug code for later, so users can help track crystal skull mis-teleports
added sanity checking to various boomerang lua files, since they assumed the controlForce parameter was always set in projectile config

Kherae:
commented out some debug code.
fixed FU tricorder stat window (oops)

Kherae:
added: removed incorrect code in monster_primary that was causing errors, and replaced it with code that should actually do something. specifically relates to hunting drops.

Kherae:
added:
fixed bow animation throwing asset errors because the thing wrongfully has fullbright information

PR #2196, opened 2020-05-05
bury axe deeper

Kherae:
put further axes into biomeairless warnings to make it stop getting up so much. Clearly it needs more axes.


Kherae:
added:
precursor data key is now a usable tool to control specific objects.
precursortinyswitchkey and precursortinybuttonkey added. they are controlled only by the data key.
void warp locus now functions in player stations and asteroid fields. also functions anywhere if admin (as seemingly intended)

Kherae:
added:
corrected precursor tiny switch series alignment and space usage. it now takes up 2 spaces instead of 3.

PR #2197, opened 2020-05-06
pump fix

Kherae:
(hopefully) fixed an issue with liquid pump, (pressurized) output pump involving stuck protection states and catch-22s


Kherae:
added:
added item sensor and item dropper, unlocked from silicon boards and iron bars
added sggunstaff and combo pistol to item network category config
moved timer gates in sunset riders to where players can't cheese the encounters by setting them to 1s

Kherae:
added:
fixed a 'not' that was where it was not supposed to be, resulting in foods having unused rot values that they shouldnt and preventing foods from rotting that should

PR #2198, opened 2020-05-07
stuff

Kherae:
new tool: point sensor. detects placement point of objects within 20 tiles. unlocked by picking up a storage bridge, crafted in nanofabricator
fixed a script error in the marvin code
fixed warped regulators and modified pillars running too fast and losing effects faster


Kherae:
added:
added missing text for some medical station dialog
updated medical station effects to use fu_generic series. adjusted scaling on hivemind. damage is now 5 percent of health, adjusted by half of modifiers to powerMultiplier. also made it use nicer sprites, gave it tracking and reduced its speed, and it now applies beesting

PR #2207, opened 2020-05-14
stuff

Kherae:
fixed goof in elder set bonus that could cause errors on first update
added special effect to roadkill augment. Meow.


Kherae:
added:
relocated and renamed all crucible UI files to 'fuweaponupgrade' instead of 'weaponupgrade2'

PR #2209, opened 2020-05-15
cat crash

Kherae:
added proper caching system to set bonus code
added keyed precursor buttons and switches to loot
added and corrected tags in too many files to list. many set weapons were lacking their set tags, etc.
fixed IDs on titanium, tungsten, and effigium broadswords and longswords. and now there's an effigium broadsword.


Kherae:
added:
crucible no longer increases fire rate or attack speed. gave it an ever so slight  increase to base damage

PR #2210, opened 2020-05-15
fix

Kherae:
oops


Kherae:
added:
replaced 'mininglaser' tag on all non mining lasers with upgradeableTool. added that tag to mining lasers. adjusted multiupgrade UI accordingly. (this requires the PR in FR as well due to implementation)

Kherae:
added: 
altar of lost souls object. crafting interface still required.

PR #2211, opened 2020-05-15
sorting

Kherae:
minor PR.
changed vanilla raygun to 'energy' category, so it sorts into weapons with bk3k.
added test ops to categories.config.patch. also reformatted.
sorted player.config.patch. reformatted to look cleaner.


Kherae:
added: 
moved autohealer series wiring over one space, so that it is now on the sprite.

PR #2212, opened 2020-05-16
fixes

Kherae:
got some leftovers that I missed from the crit changes...and fixed the critchance1/2 effects to properly give crit chance. added critdamage1 to give critdamage
fixed an incorrect unlock for dreamer legs


Kherae:
added:
fixed aetherium scythe alt erroring, potentially more weapons using futravelingslash (fixes error. if they dont fire sound, it isnt configured right)

PR #2249, opened 2020-06-13
 corrected some bee shortdescriptions. Fixed typo on bee codex. Changed type of copper ladder

kukagres:


sayterdarkwynd:
cool, though the descriptions were likely pointless, as those are set in the bee script. But , useful none the less :)

PR #2263, opened 2020-06-24
Rework Advanced Liquid Sensor

B-Reif:
This PR reworks the Advanced Liquid Sensor. With these changes, the sensor has a smaller wiring footprint, and more powerful detection.

- The sensor now has a single container slot.
- The sensor now has only one wire output (was 9 outputs).
- The sensor's script now compares the liquid at its position to the item placed in its container slot.

Some issues:
- This will break existing setups using the Advanced Liquid Sensor. Any existing setups will not output any signal.
- I'm not certain how existing setups would respond to the sensor having fewer outputs than before. Would this cause a crash? Would the wires be removed silently?

If backward compatibility is a substantial issue, we could move the new implementation to a new item. The prior implementation is very limited in usefulness, so I doubt many setups exist at this point anyway.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@ Desktop.ini
 *.sublime-project
 *.sublime-workspace
 
+# VS Code settings
+.vscode
+
 # Recycle Bin used on file shares
 $RECYCLE.BIN/`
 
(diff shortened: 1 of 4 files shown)


B-Reif:
Using this you can make a liquid router:

image

PR #2265, opened 2020-06-25
More Planet Info Cultist Lair Patch

rylasasin:
Automatically patches More Planet Info to include Cultist Lairs... if they have MPI installed.


rylasasin:
... Actually, if we could get permission to fully integrate [More Planet Info Lush Planet fix](https://steamcommunity.com/sharedfiles/filedetails/?id=1761193918) while we're at it, that would be great.

rylasasin:
I'm going to add to this, before I push it. Canceling to prevent premature pushing.

PR #2267, opened 2020-06-26
Cultists and Wrexm

rylasasin:
Two things with this.

1. Since we're using the cultist types for more than just the Hylotl mission now (those types are used for the Glitch Castle mission, the Cultist Lair, and the Cultist Ship, and if I get around to it, for the last bounty mission too,) I thought I'd give them more leveled loadouts. Might be overbearing, let me know.

2. Finally decided to upload the wrexm nar space encounter version I have been working on. Works well enough when spawned in through placedungeon, but I have yet to test it in its native enviornment (as a space encounter). It's just the planetside version that's been adopted to a space encounter variant. The bottom hatch is now approachable with a mech, and there's turrets, plus the shield generator is gone (might mess things up), but otherwise the encounter is identical to the planetside one. Oh, and there's (supposed to be) a secret way in now too.


sayterdarkwynd:
looks good. we'll need to keep an eye on the cultist damage output at tier 6 though lol

PR #2273, opened 2020-06-30
Microdungeons and Randomized Spawns.

rylasasin:
- Readjusted space enemy groups so they now use the spawner. Might need further adjustment. Basically, enemy groups are now completely random instead of a bunch of set groups. You never know what you will encounter (nor how many) in any space area now.
- Converted a few microdungeons from my FFS Dungeons and Encounters mod. Namely, the Cultist/Pirate (now bandit) tents, and the Bandit Tower. They have randomized spawners. I don't know how you want to distribute those, so I'll leave that up to you.
- New Space Encounter: Elite Bandit Asteroid Base. Another conversion from my FFS mod, this time a much heavier makeover. They are the same types of FU Bandits that appear on Eden worlds. There's a wrexem boss in the first floor to replace the pirate boss. He's a placeholder until we can come up with another monster/npc to replace him.
- Removed "wastelandquests" 'dungeon' from rotation. It no longer serves any purpose since you can get the questgiver's quest from the science outpost.


rylasasin:
**SPECIAL NOTICE:**

If, while testing, you come across the following error in your log file:

`exception caught: (MapException) Key 'nada' not found in Map::get()`

_Do not,_ I repeat, **DO NOT** report or attempt to fix this. _**This is 100% intentional.**_ The "error" is actually a feature with randomized spawners that allow a random amount of enemies to be spawned via certain spawner.
IE: In the "camp" microdungeons, you may be attacked by anywhere from 2 to 14 enemies, the exact number is completely random.

To accomplish this, I have one of the npctype/monstertype options to pick be "nada", or, nothing. When it picks this, it will throw that error, and not spawn that monster, and move onto the next one.

IE: It's a bit like the "chance of none" feature in Fallout leveled lists.

sayterdarkwynd:
You shouldn't ever be "causing" errors to create features. You should instead create a print line in the lua file to create a statement in the log.

rylasasin:
I was trying to create a method that didn't rely on lua screwery, let alone creating a whole separate lua file for it. (I'm not even sure how exactly to do what you're proposing, tbh). 

14 is just the absolute max, though actually getting 14 bandits in that encounter is _extremely_ unlikely. I neglected to mention that in that particular encounter, each bandit (outside of the first two) only has a 1 in 5 chance of spawning, so you usually end up with somewhere between 3-7 bandits.

PR #2295, opened 2020-07-12
furnace and derp

Kherae:
adjusted furnace code.  it now will run if there is a bonus OR normal output, and can accept tables or strings for normal outputs.
further update to bows: removed vanilla bow/energybow anim file changes entirely. switched over to FU specific variants. FU and vanilla bows now use these variants
removed bow overrides from vanilla harpoon gun, as these were causing errors. implementing properly will require  spriting and code work far beyond my current edits' scope.
touched up on memento mori: skull tooltip text now fits fully. also, it won't fire death effects for admins who take 'lethal' damage anymore. added benefit of admins not losing med station buffs that way.
touched up crew contract tooltips to make them fit better. gave them a bit more room to fit in.
did the same for killpods.


sayterdarkwynd:
i already made the madnessvoxel fix. you can remove that.

PR #2331, opened 2020-08-06
Fix Atmospheric Condenser

Vollch:
Atmospheric Condenser relies on "ship.celestial_type" property to track current planet when installed on ship. "flyShip" it's wrapper to "celestial.flyShip" function which set that property. For some reason it was broken in 089a848cc5a69122cc4092573d65ea20662a0e84.

### DIFF
--- a/interface/cockpit/cockpit.lua
+++ b/interface/cockpit/cockpit.lua
@@ -1281,7 +1281,7 @@ function planetScreenState(planet)
             orbiting = shipLocation[2]
           end
           if orbiting and compare(coordinatePlanet(orbiting), coordinatePlanet(self.travel.target[2])) then
-            celestial.flyShip(self.travel.system, self.travel.target)
+            flyShip(self.travel.system, self.travel.target)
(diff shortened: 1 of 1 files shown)


sayterdarkwynd:
What about this is broken? Not sure what this is doing. Been a long time since I touched this.

Vollch:
Condenser, when it build on ship, supposed to gather resources of planet where ship currently orbited. But since objects doesn't have access to celestial api it's tracked here, in cockpit script, via "ship.celestial_type" property. Without propper tracking output will(Actually, since 2 of 3 calls to celestial.flyShip() still wrapped - sometimes it works, but very unrealible) fallback to default table no matter of where ship actually is. 

Here's two years old PR where such mechanic was added: c82c0d59e884a6447819b41345d64e56d648e224

sayterdarkwynd:
thanks , that clears it up. approved.

wanted to be sure. I literally have 0 recollection of ever touching this, but it clearly states that i did :)

PR #2338, opened 2020-08-09
Fixed embalming brains

Vollch:
With recipes "item a = item x", "Item a + item b = item y" and "item a + item b + item c = item z", having all three items(a, b, c) in machine we can't predict which item(x, y or z), we'll get in result. Pairs() in map function doesn't respect order of elements, and causes mess in table of recipes, and startCrafting() just takes the first one from the top of messed pile. Your mileage may vary, but in my case having all 3 items i was always used only 2 of them.

In embalming table we have a collision of three recipes for brains of different tiers, now recipes will always be proccessed from top to bottom. First one which matches given items - will be performed. That's why they also was reordered, moving more prioritized recipes upper.

I've seen a note in extractor code regarding same issue, and some commented out recipes. It can be reenabled as well by adopting same fix, if needed.


Vollch:
Also, with 2nd commit Lab Directory now show the recipes of autopsy table.

PR #2339, opened 2020-08-11
purposefully mysterious farts

Kherae:
Misc fixes:
updated bag of farts to be mysterious  again
added a sanity check to burrower death projectiles
further fixes towards energy related errors on melee weapons
fixed invalid item ID on crewmembermedic2
corrected duplicate black slime hat in a collection.
corrected macready, fueyebeast, and mutantminer having 10x the protection they should have, resulting in them taking almost no damage due to previous changes in Defense's Damage Reduction scaling, which was previously capped almost 10x higher.
halved maggot health.
corrected 4 greataxes to use greataxesmash instead of hammersmash
reduced query area of a specific checkpoint in the lunar base, which could result in players getting stuck if they died without opening nearby doors.
corrected pony modpack raceeffect files. again. 
pony modpack race descriptions shortened to fit the tooltip on the char select screen.

reworked rimeling:
ice field status effect recoded to properly slow enemies in range. min range on big area sphere removed. big sphere time to live is now 6s, matching the cooldown. all of the rimeling's spawned ice gas projectiles last 4s at most. ice gas projectile no longer ignores defense. ice gas spawned by big orb had its damage inheritance reduced to 0.15, from 0.25. freeze effect now has a resistance cutoff at 75%. 

Capture Gun rework:
No longer damages or impacts non-monsters or non-capturable monsters. Can no longer bring monsters below 1 HP. Is now upgradeable. Starts at tier 1 (down from 3). Has 2 Versions now.
Current version (20m in 10k voxels) is now percentage based.
New version has previous stats of current version (flat damage), new sprite, and costs 200k pixels (in 10k voxels).


sayterdarkwynd:
revert macready and fueyebeast changes. I did not ask you to adjust those, and they are intentionally that high.

Kherae:
the macready should NOT be reverted, because then he takes 1.8% damage, which is a problem. previously he capped at 100 defense,which was 13% damage taken. that was reasonable. Currently he takes 1.8% damage, because of how the protection was implemented. See dev chat for the whole math. These changes bring these enemies to a reasonable level that is STILL over what they used to be (106.5 def at t7 is 11.8% damage taken).

Summary:
Pre-Defense Scaling: Macready takes 13% Damage.
Post Defense Scaling: Macready takes 1.8% Damage
Post Khe Nerf: Macready takes 11.8% Damage

PR #2349, opened 2020-08-24
In which I provide the changes I made in #gamestart-revision.

anoobindisguise:
In summary:
- Vork the Penguin's shop lets you sequence break progression a little too hard. Instead of selling useful things like gasses, mech parts, crystals and silicon, he now sells junk ship parts which you can use to cheaply create a larger BYOS ship earlier in the game than you would otherwise.
- The starting BYOS ship is now made of junk (rusty sloped panels replace the ship's hull, while the inside is replaced with rusty metal latticed by scrap panels. 
- Completing the "build a Small FTL Drive" quest now gives you clean sloped hull panels, wall panels, ship supports and metal railings (thus giving you the blueprints) so that you can build a clean, shiny BYOS ship if you put the effort in.

### DIFF
--- a/objects/crafting/armory/armory.object
+++ b/objects/crafting/armory/armory.object
@@ -219,7 +219,8 @@
           "riflequietusfu",
           "rifleferoziumfu",
           "rifleaegisaltfu",
-          "rifleferoziumfu"                      
+	  "riflevioliumfu",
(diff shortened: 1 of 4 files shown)


sayterdarkwynd:
all seems good to me.

PR #2352, opened 2020-08-28
Tons of armor fixes

CyanSkyKnight:
First of all, 85 hats and helmets now have proper masks. The affected hats and helmets will no longer remove beaks, beards, and Saturnian eye colors (among other things)

Second, I also gave some of the helmets icons they were missing.

Finally, I trimmed some of the unnecessarily large whitespace from some armor sets, like the Recon sets.


sayterdarkwynd:
nice , thanks. This wont be merged till Research is further along but will definitely be merged :)

PR #2355, opened 2020-08-30
Ansible network price correction

hugojmaia:
No more overly cheap seeds and sanity priced ores


sayterdarkwynd:
this can wait till after Research is done. We may be dropping this station completely.

PR #2371, opened 2020-09-08
Change Data Server Icon

anoobindisguise:
It looks identical to advanced computer, which looks ugly in the research tree. So I changed it to the astro server icon (which the data server resembles almost exactly)


anoobindisguise:
I also removed the module requirement from drills. They're a temporary mining solution which are quickly outclassed by mining lasers, but every time you make one you are set back because you're wasting modules on drills. If you don't need modules to make them, then there's at least a small window where you will use drills before you move on to (infinite durability) mining lasers or (just way better) frackin' universe pickaxes - carbon, plasmic, etc.

PR #2372, opened 2020-09-09
Engineering and some stuff.

anoobindisguise:
Made a tree for engineering. It's all the tinkering, ship crafting, mech crafting, and machining table recipes.
Spaced out ship upgrades a bit (mk1abc drives are titanium instead of tungsten, mk2abc are durasteel, mk3abc is densealloy, human is durasteel, human2 is densealloy). This is because the Small FTL Drive is obsoleted by MK1s in the same resource tier (tungsten) and even if they're separated by nodes, I want the tiers to progress more.
Crappy beds are iron tier now and can be made at machining tables. 
Added a Core Fragment-based STL Drive and a red core fragment booster for traveling in your home system. The crappy mech we discussed will be worked on, eventually.


anoobindisguise:
Also, I made chem lab carbon much cheaper to craft and craftable at tier 1, since carbon is tier 1 in every other source. The initial recipe (20 co2????) was honestly kind of insulting.

anoobindisguise:
Added a chem lab + medicine tree.

PR #2482, opened 2020-10-08
Added "Mantizi Construction" node to Mantizi starting recipes

Zarra-0:
Added all of the recipes unlocked by the craftmanship node "Mantizi Construction" to the Mantizi's default recipes.


kukagres:
I think you forgot the comma at the end of each item.

sayterdarkwynd:
indeed. Zarra, you forgot them magical commas.

Zarra-0:
Fixed, thanks.

PR #2540, opened 2020-10-20
Added the Mech Platform to the mech research branch

Zarra-0:
The mech platform was added by CF in the same way that most objects were - through the pixel printer, scanned in the world. Unlocking it through the research system will incentivise players to use their mechs more, and thus continue with the mech research branch, as well as making mechs easier to use and less distinct from the rest of the game.

### DIFF
--- a/recipes/prototyper/labs/tools/mechplatform.recipe
+++ b/recipes/prototyper/labs/tools/mechplatform.recipe
@@ -1,0 +1,14 @@
+{
+  "input" : [
+	{ "item" : "titaniumbar", "count" : 5 },
+	{ "item" : "electromagnet", "count" : 2 },
+	{ "item" : "powercore", "count" : 1 },
+	{ "item" : "upgrademodule", "count" : 1 }
+  ],
+  "output" : {
+    "item" : "mechplatform",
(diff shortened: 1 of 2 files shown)


sayterdarkwynd:
i have absolutely no interest in letting players summon mechs any time prior to completing the bounty system. Its cheap, breaks every single boss fight, etc.

sayterdarkwynd:
also..conflicts here.

Zarra-0:
The only boss fight I can think of that has enough non-protected terrain to place down the spawner is Shog, which is late-game enough that players could have completed the bounty system at that stage anyways. I'm not too fussed over it, though. Just wish I had the option to bring in my mech halfway through exploring a world.

I don't understand the conflict report. It looks like the regular "add a word" change that's merged just fine in the past.

sayterdarkwynd:
well, ill give it some thought. Not hyped about the idea, but if its just the vanilla object I'm not overly concerned.

Kherae:
if we do add it, we need to make shoggoth one-shot mechs with an eyebeam laser

sayterdarkwynd:
added in, manually

PR #2586, opened 2020-10-31
Fix comments

rainy-goose:
collections row comments didn't line up with what is shown in-game, this is due to row 1 having 3 collection tabs


sayterdarkwynd:
collections are currently in the process of being removed from the mod and added to a separate addon. this can go into that.

PR #2619, opened 2020-11-11
Added recipe for the Energy Pickaxe to the Mining 2 node

Zarra-0:
Energy pick was a uselessly rare vanilla drop, that most users would never find before upgrading their MM to the point where it becomes unnecessary. Adding it to the research tree lets it fill a niche for base-building and general handiwork before the user gets the morphite manipulator, while still being significantly weaker than the (unlocked earlier) mining laser and other available pickaxes/drills.


sayterdarkwynd:
Yea. No thanks. This should stay a rare drop.

PR #2648, opened 2020-11-23
Spirit Tree Race compatibility

rylasasin:
FR Support for Spirit Tree. They have the exact same stats as Spirit Guardian (Since the former is a remake of the latter)


Tigrex-Dai:
Actually we are not the same^ ^
Sorry for add support such late :)

rylasasin:
Well, if your team is going to add yours, I'll close this one then.

PR #2654, opened 2020-11-24
fishhbowl helmet

Kherae:
minor performance optimizations to some scripts
fixed veilendrex spritesheet having bleedover from indicator spaces
tech swap madness effect can no longer occur in instance worlds due to the potential for players to get stuck out of bounds
removed massive amount of trailing spaces and tabs from lua files
added checks to some homing projectile scripts to ensure they only home in on targets they can damage

power system optimization and attempted bugfix:
reduced disk read/write performance impact of objects using fupower.lua by shifting entity lists from storage to self. given that IDs are only kept while an object is loaded, serializing them to disk is a pointless waste.
implemented failsafe entity list refresh in power conduit scripts and fupower.lua in case a messaged entity no longer exist, to give the system a jumpstart.

unifiedgravmod.lua, swimming, and flight techs:
implemented a stat in unifiedGravMod for a specific flight mode override
implemented status effect to handle flight mode override for techs
edited relevant flight techs to use
revised swimming code to play nicer with FU gravity effects
revised mountain gravity and cloudblocks to use unifiedGravMod


sayterdarkwynd:
conflicts to fix

PR #2656, opened 2020-11-25
candy corn

Kherae:
apiary now properly considers environmental resistance frames
corrected trays not using colors on water use scan tooltip segment
applied fixes to unused slimepersonscanner
adjusted fission generator tooltip
completed incubator's 'forced hatch' functionality, which was never used.

nerfed toxin cloud (med station and madness variants) power to 5, instead of 50. it still ignores defense. it was hilarious doing over a hundred damage per tick, but that needs to go.

made various fixes to the glitch mission:
all cultists should now head left, instead of sitting around at the camp to the far right
cultists spawned during the boss phase will no longer spawn if there are more than 20 already active
all cultists will not properly despawn shortly after the boss dies.
fixed an issue that caused waves to progress upon killing the cultists' pets, when they were supposed to wait til all the cultists died.


sayterdarkwynd:
You're gonna need to remove the generator object file changes. ive been actively working on those and now they conflict.

PR #2670, opened 2020-11-28
Add Spirit Tree Race support

Tigrex-Dai:
add Spirit Tree Race in Tricoder and meta file.


sayterdarkwynd:
you should probably fix the scripts you made first. Numerous users reporting errors.

Tigrex-Dai:
> you should probably fix the scripts you made first. Numerous users reporting errors. I suspect you didnt test as well as you should, as they are saying energy doesnt regen, etc.

Yes I am trying to fix, but there exist errors including files of Spirit Guardian race and idk why. Am looking into the problem.

Tigrex-Dai:
> you should probably fix the scripts you made first. Numerous users reporting errors. I suspect you didnt test as well as you should, as they are saying energy doesnt regen, etc.

I think I should cancel the day&night effect first, and put it back when the problem is solved

Tigrex-Dai:
Emergency fix, cancel effects that will cause problems and will reactived when problems are fixed

PR #2693, opened 2020-12-04
Fixed Sail disconnecting player and added BYOS support. (Spirittree)

zombiex227:
Accessing S.A.I.L as the Spirit tree race caused an incoming client packet exception. This change fixes that, as well as adds BYOS support for said race.


sayterdarkwynd:
please remove the metadata edit. Thats something only I edit , for the exact reason its currently displaying: conflicts.

PR #2694, opened 2020-12-04
Switching which Sergal mod is used.

KH007:
LangyX's branch of the Sergal mod is more up-to-date than the other ones.


sayterdarkwynd:
sure, whatever. Still the same creepy-assed race either way.

PR #2697, opened 2020-12-05
ugh

Kherae:
spirit tree race:
fixed FU's includes
revised tooltip
changed wand damage bonus to multiplicative rather than additive, reduced bonus to 20%
streamlined day and night bonuses

xi bulbs:
corrected stacking behavior for food drain, between multiple tiers of plant.


Tigrex-Dai:
Thx a lot^ ^

PR #2699, opened 2020-12-06
Updated Shoggoth mod name

Runchimchar:
Had included "The Shoggoth Race Mod" instead of "shoggothracemod", which caused stats to not always display in tricorder


sayterdarkwynd:
conflict. ill manually do this.

PR #2719, opened 2020-12-11
Fix a buncha typos, add wine cabinet

Zennle:
consume typos for sustenance


sayterdarkwynd:
to be clear, adding periods is not "fixing typos". It's nitpicking. The periods were omitted intentionally because they are entirely irrelevant on tooltips

PR #2739, opened 2020-12-14
Typo + style consistency fixes

danuguy:
Now in a single pull request. Feels like there's still room for improvement in my workflow though


Zarra-0:
The "An electromagnet" change to fu_quests.radiomessages is incorrect. The portion of a sentence following a colon is still part of said sentence, and so shouldn't start with a capital.

PR #2760, opened 2020-12-21
quack

Kherae:
replaced debug code in FR liquideffect script with a failsafe that will assume defaults from player.config, if the npc lacks a configured mouthPosition status property.

revised particle emitter in FR perfect block bonus code, due to compatibility. will now check that the item's animator supports before attempting to play, and will attempt to derive an alternate if it does not.

fu_genericStatApplier should hopefully no longer error due to missing stat modifier group handles. instead, it'll print what effects are active to aid in tracing the issue.
fixed darkcommander causing issues with the above

added missing metamaterial from shreddy's blocks


sayterdarkwynd:
I do not want the particle removed from shields. Put it back.

Kherae:
the particle emitter was removed due to causing weapons that DO NOT have it crashing. I left in the sound, for weapons that support that, as at least we have animator.hasSound (no similar function for particle emitters). I wouldn't be removing it if it weren't causing issues.  At least it isn't a repeat of shortswords this time.

sayterdarkwynd:
That's nice. It stays. Other mods that are not playing nice are, quite honestly, not my problem. They can update on their end to fix the issue. All vanilla shields and FU shields work fine.

Kherae:
revisions made.

PR #2800, opened 2021-01-02
Update isn_thermalgenerator.object

Edulml:
Rebalanced the burn time of unrefined wood (fullwood2). It will still be beneficial to process the log into it, but a single log processed into 10 of these for a total of 50 seconds will no longer be the second best fuel in this machine.


sayterdarkwynd:
would rather see this a bit higher. You're telling me paper burns better than wood in this thing? Perhaps rate it at 3, not 1.

PR #2803, opened 2021-01-03
hylotl breath

Kherae:
fixed hylotl water breath bonus: reduces breath depletion rate, rather than increasing max breath, since the latter doesnt work as was expected

### DIFF
--- a/species/hylotl.raceeffect
+++ b/species/hylotl.raceeffect
@@ -112,7 +112,7 @@
 				{ "stat": "maxHealth", "effectiveMultiplier": 1.25 },
 				{ "stat": "physicalResistance", "amount": 0.2 },
 				{ "stat": "perfectBlockLimit", "amount": 2 },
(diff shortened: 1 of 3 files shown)


sayterdarkwynd:
It worked perfectly fine before. I fail to see how this would suddenly change. Not merging this until I have time to check it out myself. I don't -want- breath reduction rate, i want max breath first and foremost.

Kherae:
Operating based on this:
7:11 AM] Edward Chernenko:
I tested a new Hylotl character, and they indeed don't get a longer breathing time (tested in Salt Water ocean). This appears to be a bug. I did get a health bonus when in Salt Water, though.


Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

sayterdarkwynd:
just tested. Latter works *exactly* as scripted. No change needed.

PR #2805, opened 2021-01-03
Update create_clothingfabricator4.questtemplate

Argenttip:
Typofix


sayterdarkwynd:
why are you removing the conversation? no thanks. Remove that change please. I'd rather give them unique text.

PR #2810, opened 2021-01-04
Tenant re-organizing

AlienAtSystem:
Changes

- moved skath tenant definitions to tenant folder (were in npc folder), integrated skath random villager patch into fu patch
- cleaned up ordering of tenant folder, going by species foremost
- removed duplicate elduukhar villager definition
- removed "fu" tenant for being a duplicate of vanilla scientist
- removed "villager_futenant" because the underlying tag ("fu") is confused and for being a duplicate of vanilla scientist


sayterdarkwynd:
not sure what most of this is even being done for?

"removed "villager_futenant" because the underlying tag ("fu") is confused and for being a duplicate of vanilla scientist"

we do not just arbitrarily remove things. That tag is likely there for a reason.

AlienAtSystem:
I plan on adding a few tenants, so this is basically a first "good foundation" pass over the files to make sure there's no legacy weirdness. (Like for example some tenants being defined in the npcs folder and some being defined in the tenants folder).

The "fu" tag might have originally been meant for the FU-specific crafting stations, but now also includes several furniture sets with little connection amongst each other (Aether, Aen and Cosmic might sort of fit together, but Cute is certainly not, and none of those fit to the FU crafting station style. And all four feel like they should spawn something different than a vanilla human scientist). I don't know the logic behind whoever tagged the items, so the less impactful change to me is to remove the tenant instead of untagging all those items.

The purpose of specific tenant types above the basic tenants is to provide some variety and sense of discovery in putting interesting furniture into a room and seeing someone unique pop up. Neither the frackin.tenant nor the villager_fu.tenant provide that, because they are boring in result and their low requirements plus high-ish priority actually get in the way of spawning interesting tenants (this is speaking from experience. It once got a colony entirely filled with human scientists simply because of the _doors_ providing enough of the science tag for the frackin.tenant, independent of what I put in the room).

sayterdarkwynd:
thats all well and good, but I will not be approving any change where you remote assets or tenants. This is not on our agenda, and many players will be using that tenant already. Removing it from them is a dick move.

beyond that, im totally fine with more tenants and the like. Just avoid removing things wholesale without our asking.

AlienAtSystem:
I am willing to compromise as follows:

- I restore villager_fu.tenant but make them look different from a normal scientist
- frackin.tenant is deleted for being a duplicate of the vanilla scientist (If you need further convincing, it currently completely blocks the vanilla hazmat tenant from spawning, who also needs the science tag but only has priority 4)
- I don't think there's any discussion about the deletion of the duplicate elduukhar definition

PR #2815, opened 2021-01-05
Minor fix.

Edulml:
Fixes quest text: this is now done at the machining table, not a ship component assembler.


Edulml:
Nevermind, I just forgot about the research node with the component assembler.

PR #2829, opened 2021-01-10
 Added Small Airlock Rail Hatch[outpostairlockrailhatchsmall] object and recipe.  Unlock in research tree is missing.

kukagres:


sayterdarkwynd:
can you please add it to the research tree too?

PR #2832, opened 2021-01-10
madness tweaks

Kherae:
added chance for positive events for some madness effects:
chance that you stab yourself and gain regeneration for a short duration
chance to gain a boost to armor (of varying strength), health or energy
chance to consume a molotov, igniting all nearby enemies for a very short duration
chance to not need to breathe
chance to gain a boost to research gain for a duration
chance to become invulnerable for a very short duration
chance to gain a boost to charisma
chance to become well fed

adjusted station buy/sell modifier adjustment parameters

fixed nightar vest using nonexistent tooltipKind


sayterdarkwynd:
the only one i directly have issues with is the igniting enemies one. Explain to me how being crazy would set your enemies on fire, exactly? I dislike that a lot.

Kherae:
I left it purposefully vague, much like the one where you somehow light your own seat on fire which it is paired with:
"Well, it would appear that your current seat has randomly caught ^orange;on fire^reset;. That, or you set it on fire. Either way: it is ^orange;on fire^reset;. And so are you."
we can pretend the player decided to throw something flammable. hell, we can make it require and consume a molotov if you want. that actually wouldnt be such a bad idea.

PR #2903, opened 2021-02-04
Elder Tenant update

AlienAtSystem:
Tenants spawned from solely the 'elder' colony tag are no longer merchants, but villager-type. They have a greater selection of clothing to wear and now also use the R'lyehian dialogue of the Temple Builder tenant.
Temple Builder now buys things for the usual 20% of value. He's already fleecing people enough with per block prices of a hundred, no need to add insult to injury by forcing people to get a second merchant for selling purposes. The dialogue of the Temple Builder has been extended to no longer have "unintelligible gibberish" entries.
Added a new tenant, Cult Artist as a second elder merchant. The artist requires the 'madness' tag in addition to the 'elder' tag, and sells a variety of art-related items, including raw materials as well as finished books, paintings and sculptures. They have a chance of selling elder relics including the Elder Artifact (this with an extremely rare chance). The artist also uses the R'lyehian dialogue.
Also added colony tags to one elder statue that was still missing them.

I am open to discussions about the chances of items in the Artist's merchant pool.


sayterdarkwynd:
Im not sure how I feel about literally any of this.

Firstly, nobody cleared this with me and its nowhere on my agenda to allow players a smorgasborg of elder tenants.

Second, I never intended to allow easy access to any of the resources. Absolutely -NO WAY- to literally almost all of the goods you mention are sold. The user is intended to visit elder temples or do madness to get access to it. Artifacts, for sale? Absolutely not.

AlienAtSystem:
I am adding _one_ elder tenant, bringing the number from three to four. This is not more than available for any other race.

In addition, as it stands right now, the fucultist tenant is painfully clearly unfinished, being labelled "Chef" and using generic Chef dialogue, but selling standard merchant inventory. This is immersion-breaking to anyone summoning tenants using elder furniture.

I can understand wanting certain resources to be rarer, but especially with the elder artefacts, I would argue the current way is going about it the wrong way. The artefacts are dropped by a _single_ mob, spawning in a _single_ sub-biome, found reliably only on a handful of planet types. It funnels exploration down into a single binary decision to be made on every planet, sapping all joy out of it.

Unlike Manipulator Modules &co, which can be found everywhere in small quantities, the elder relics to do not reward thorough exploration. You find a planet with the right mob, plant a flag and then just farm those mobs night after night. It's boring and repetitive. And because only one drop table is involved, the player is soon swamped in very specific items that they have to throw away constantly. Also note that this includes the paintings which you condemn here for availability.

Also, even if the player gains access to some elder relics via the artist? These items don't have any obvious use. Their recipes aren't even in the wiki. People still have to find codices and also importantly, still finish the madness tree to gain access to the Elder Altar to actually have use for those items. But to somebody who has the required prerequisites, it can be a bit of a relief to be able to throw pixels at a problem instead of having to farm another night in the elder biome for that one thing they need.
And that is of course assuming the merchant sells the right artefact in the first place. To have a full set of artists selling all normal relics, a player needs to summon on average 130 elder artist tenants. To gain access to the Elder Artifact, they need to summon a 1000.

I don't follow the argument about "park their arse in front of it". People are already parking their arse on Atropus planets to farm madness without going up the tree (which happens because the tooltips for Insane Notes &co don't mention that these are consumables giving madness), and in addition I'd like to point out the Artist tenant requires the "madness" colony tag, which means a crafting station from the madness tree. By the time you can gain access to the artist, you can make Insane Notes by yourself already.

sayterdarkwynd:
The Chef is honestly probably fine. It's the rest I took issue with.

"The relics are dropped by a single mob, spawning in a single sub-biome, found reliably only on a handful of planet types"  

yes. That's on purpose because they are ONLY intended to be found in Delta Freya. The player is not *intended* to be likely to find lots of them. They are (players) not guaranteed to ever get Elder stuff in a playthrough. This is how they were designed.

I am not ever allowing easy access to the Elder Artifact. There is no argument to be made here that I will agree with on this front. 

The rest can be argued for to some degree. The tenant being difficult to get, that's great. No objections with that. Lower tier art is fine, as well. Not high-end ones though. 

In any case, I'll merge in and simply change the bits I don't like. I am not interesting in arguing with you on what you feel is right with -my- project.

AlienAtSystem:
I think we misunderstood each other: The whole "Chef" thing was the case before the commit. Now the standard elder tenant is a villager-type instead (not selling anything), and the artist as a new one sells non-generic things. If you want to adjust what the artist sells, that is of course your prerogative.

PR #2907, opened 2021-02-05
energy regen zone nerf

Kherae:
changed energy zone's energy block time multiplier from 0 to 0.05.
removed blocking effects from neuropod, as it no longer applies energy regen effects.


sayterdarkwynd:
why though? This will make things with energy zone nearly worthless. Please run around with the lasrifle a bit and make sure the alt doesnt suck shit now? If its still good, I'm willing to try this.

Kherae:
revised to 0.125. with 0.125, you can stop firing for a brief moment to get a big chunk of energy back on the lasrifle.

main reason for the nerf is that as 0 it enables certain techs and such to be infinite use. that's how that one pair managed infinite duration phase attack.

PR #2911, opened 2021-02-06
Minor Update to cadaver.item

AnomNum:
Adjusted name and description.


sayterdarkwynd:
rather needless, but whatever ::)

PR #2914, opened 2021-02-07
Tilesets

Zarra-0:
- Increased the range and effect duration of the BYOS gravity generator, for easier shipbuilding.
- Replaced the platforms in the Thelusian BYOS template with Ruins platforms.
- Added a Skath BYOS template.
- Added a Mantizi BYOS template.
- The Precursor Forge now appears in the Lab Directory.
- Reduced the pixel value of the lab directory.
- Added five new Mantizi ship greebles.
- Mantizi banners are now gilded once more.

Gilded images from #contributions [here](https://discord.com/channels/216411946017095693/336895240453619712/799815924805795871) - they don't seem to have ever PRed it


sayterdarkwynd:
why does the acid EPP provide oxygen? No thanks. It not having oxygen is 100% intentional, so that the player is forced to make a choice ::  either craft the EPP and an O2 capable suit, or craft the Acid Proof suit and an O2 pack.

Zarra-0:
> why does the acid EPP provide oxygen? No thanks. It not having oxygen is 100% intentional, so that the player is forced to make a choice :: either craft the EPP and an O2 capable suit, or craft the Acid Proof suit and an O2 pack.

I don't see any reason why a player would craft an Acid EPP and a rebreather/oxygen-providing armour set, instead of using a regular airtank, which they would have anyways from exploring their first moon, with acid-proof armour or an acid augment.

Especially since the Acid EPP is available later (requires zerchesium) than Battleborn (requires carbon plates) or Acid Shield (requires advanced circuits/advanced plastic)

sayterdarkwynd:
none the less, im not adding that change to the acid pack. I gave the reasoning.

PR #2925, opened 2021-02-14
Selection of Slime racial scan text for vanilla objects

bioboygamer:
### An admittedly eclectic group of scan text for vanilla objects for the Slime race. 

Since I know Sayter's busy, I'll make this short(ish). 

The scan text generally follows the current Slime lore, and adds to their characterization as **curious and philosophical**.
It portrays a Slime player character as a **contemplative and very well-read Protectorate agent**, with perhaps a lack of genuine personal experience and some emotional issues that are the logical result of being born from a dead body.
Notably, the current Slime racial furniture and ship part scan descriptions portray the Slime race as unintelligent, which contradicts their usual NPC dialogue. I'm erring on the side of assuming higher intelligence, given the way they're generally described throughout the mod.

### Minor notes (For anyone who really cares):
There may be some minor inconsistencies between these and the Slime NPC Dialogue. To tell the truth, I had completely forgotten about NPC dialogue until after I'd written most of the descriptions. Luckily, there aren't too many irregularities. NPC Slimes talk much more casually, which could be chalked up to a natural difference in personality, or just not necessarily being a Protectorate agent like the player. There may also be some inconsistencies in the details of the Slime's biology, such as how sticky they are or aren't, or how they react to water, but that too could probably be a case of variation between individuals.

You might notice that the Slime's opinions on Novakid objects are quite a bit harsher than their comments on other races' objects. This is more or less intentional - It's not that the Slime(s) hate Novakids as a race or species, but the Novakids' in-the-moment lifestyle and apparent general disinterest in self-reflection is grating for a species that values introspection and learning. It's a clash of cultures more than anything else.

There are a couple of in-jokes that I couldn't hep but add; Ultimately, I won't be too broken up about it if they get axed, so if you want them out, let me know.

**TL;DR**
I made some Slime scan text for vanilla objects. They're **curious, philosophical, and love to learn**, but also have some issues. They're annoyed by how lackadaisical the Novakid tend to be. **If there's anything you don't like, let me know**. Or remove it, or just toss the whole thing - it's your mod, after all (although I'd prefer to be able to make improvements if you have issues with anything).


bioboygamer:
Also, this is very much not a big priority. Feel free to ignore it if there's anything remotely more important.

PR #2944, opened 2021-02-25
Update tungstenshield and ironshield groups

RusTit:
I've noticed small bug. "Tungsten Shield" appearing in the armor group on "Assembly Line".
I've not familiar in Starbound modding, so not fully sure, is my PR fix this, but I've checked "Lunari Shield" as example and it has all groups except "craftinganvil" and "armours", so I thought that removing "armours" group should fix the issue.

```
// recipes/armory/shield/lunarishield.recipe
"groups" : [ "armory1", "shield", "all" ]
```

### DIFF
--- a/recipes/anvil1/weapons/tier1/ironshield.recipe
+++ b/recipes/anvil1/weapons/tier1/ironshield.recipe
@@ -7,5 +7,5 @@
     "item" : "ironshield",
     "count" : 1
   },
-  "groups" : [ "craftinganvil", "armours", "all","armory1","shield" ]
+  "groups" : [ "armory1", "shield", "craftinganvil", "weapons", "all" ]
 }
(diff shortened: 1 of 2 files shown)


RusTit:
If my fix is correct, I can make another fixing commit for "Iron Shield"

Zarra-0:
It has that group so that it shows up in the vanilla anvil, which lacks a dedicated shields tab.

RusTit:
@Zarra-0 
To make it clear:
"craftinganvil" group is used to show in vanilla anvil, right? or it's "armours"

Zarra-0:
Both. `craftinganvil` is the workstation, `armours` is the tab. Instead of removing the armours tag instead replace it with "weapons", which doesn't match any Armory tab but does match the anvil weapons tab.

The vanilla basic shield recipe, for comparison:
`{
  "input" : [
    { "item" : "ironbar", "count" : 4 },
    { "item" : "darkwoodmaterial", "count" : 20 },
    { "item" : "string", "count" : 4 }
  ],
  "output" : { "item" : "startershield", "count" : 1 },
  "groups" : [ "craftinganvil", "weapons", "all" ]
}
`

RusTit:
@Zarra-0 
Thanks for this. I followed your advise and make another commit (also updated ironshield).

PR #2957, opened 2021-03-03
Update mechparts_legs.config.patch

rylasasin:
nerf the elite sigma mech legs (still better than normal sigma legs though)


sayterdarkwynd:
I am unclear why these need a change?

rylasasin:
Too OP for its tier.

PR #2959, opened 2021-03-05
brain harvester tweak

Kherae:
fixed a case where brain harvesters (via deathbombcomplextreasurepool.lua) werent picking up body material kinda due to a different configuration implementation. repair bots should now properly drop artificial brains.

corrected kirhos species description: they have no immunity to electrified, in whole or in part. They merely have 50% electric resistance (which has no impact outside reducing electrified's damage)

minor script optimizations in some effect scripts.

corrected electrobaton2's stance durations to reflect its faster fire time, thus making the weapon actually functional again.

implemented recursion limit (100 iterations) in railtramstop.lua, because some stupid troggs managed to force a recursion error.


sayterdarkwynd:
remove the kirhos edit. i already did that on my end this week

PR #2982, opened 2021-03-23
Added proof-of-concept Science Outpost Info Booth

bioboygamer:
I have no idea how to make this not include the changes that have happened to the mod since my last update. Like, how do I not have that ugly "Update" commit with a million unrelated changes as part of this pull request. I do not understand github. I am sorry.


sayterdarkwynd:
First, stash your current changes somewhere

then, pull from master into your fork/repo

that should update the PR on its own, if I recall correctly, and hopefully remove those extra files

bioboygamer:
> First, stash your current changes somewhere
> 
> then, pull from master into your fork/repo
> 
> that should update the PR on its own, if I recall correctly, and hopefully remove those extra files

I will do all of that as soon as I figure out how to do all of that, which is to say as soon as I figure out how to do the first stash part. I've been able to determine that it's something that I'll need to do through GitHub Desktop rather than just the GitHub website, but I am unsure of the specific set of actions that I need to perform in order to do this.

PR #2986, opened 2021-03-26
Removing Former Gallery and Fabricated World

Pixelflame5826:
All it does is remove them form the treasure pool, any one of them still in the game will still be there, but it's to not break any current builds.


Pixelflame5826:
Stupid me, I'll try to fix an issue.

PR #2987, opened 2021-03-26
Proof-of-concept Science Outpost Info Booth (finally)

bioboygamer:
So, this _should_ just be the necessary files for the Info Booth. As per usual, feel free to add or remove whatever you want - or ask me to do it. Huge thanks for taking the time to help me make my future pulls less awful.


objects/scienceoutpost/scienceinfobooth/scienceinfobooth.object
```diff
@@ -0,0 +1,91 @@
...
+	"Looks like most of your Protectorate crafting databanks have been scrambled at some point - You'll need to use your Tricorder to research how to craft a lot of things.",
+	"I'd recommend researching extraction technology sooner rather than later. A Hand Mill might not be as fast as a full Extraction Lab, but it's still one of the most powerful tools in your arsenal.",
+    "Having trouble keeping your ship's tank topped up? Remember, Erchius isn't the only fuel you can use - Radioactive isotopes, properly processed algae, and even pure Hydrogen can all act as fuel sources - and some are more effective than others!",
+    "Having trouble gaining research? Your passive research gain will increase the longer you play, and will also increase based on the tier of the planet you're on - Exploring more dangerous planets lets you gather research more quickly!",
+    "Okay, so a couple of coworkers were saying they were leaving to go snarsh some garflipps, and I asked them what that meant, and they looked at me like I was an idiot. Is that normal? Am I the wierd one for not knowing what that means?",
```
bioboygamer, line 71: Does it "spoil the joke" to have a character not know what snarshing garflipps is? I kind of like the idea that the player first sees the term here and assumes that it's a running nonsensical joke, which makes the eventual appearance of the Snarshing table hit even harder.

objects/scienceoutpost/scienceinfobooth/scienceinfobooth.object
```diff
@@ -0,0 +1,91 @@
...
+    "Have a favorite weapon or armor set that you don't want to leave behind as you progress? Good news - you don't have to! Using your Tricorder's Upgrade function, you can use Upgrade modules to improve almost all Frackin' Universe weapons or armor. By doing this, you can ensure that any armor or weapon you like stays viable for the entire game!",
+    "Does an enemy seem almost invincible? It might just be resistant to the damage type you're using. Try switching things up with different weapons to find a weakness!",
+    "Don't know where to find a specific item or material? A Lab Directory can tell you if it's an output from any of your nearby machines. If that fails, check the official wiki displayed on the main menu!",
+    "Even carnivores have a reason to set up plant farms - many of the growable plants that Frackin' Universe adds can be eaten by anyone. Plus, Meatvine is a reliable source of nutrition, and many plants and their seeds can be extracted into useful materials!",
+    "Sorry we can't let you use most of the crafting and processing machines - We had to restrict access after Kevin tried to use them to create a gun that shot smaller versions of itself.",
```
bioboygamer, line 77: "Gun that shoots smaller versions of itself" might as well be a placeholder - I just thought it'd be a funny explanation for why the player can't use the Science Outpost machines. If there was more room, I'd have the person at the info booth remark that they're not sure _why_ limiting access worked, since Kevin could probably bypass it, but for some reason he doesn't. I dunno. I just like the idea that they've found something that keeps Kevin in check, but logically it _shouldn't,_ and they have no clue why it works.

objects/scienceoutpost/scienceinfobooth/scienceinfobooth.object
```diff
@@ -0,0 +1,91 @@
...
+    "Most hazardous environmental effects can be prevented with either an immunity to that effect, or often through a high enough resistance to specific damage types.",
+    "Immunities and resistances are seperate - Even if you're immune to a planet's deadly heat, it won't reduce damage from fiery attacks unless you increase your heat resistance.",
+    "If things don't seem to be working right, first check to make sure you don't have any incompatible mods installed. If that doesn't work, post your full log file to the bug channel on the Frackin' Universe Discord.",
+    "Take your time - if you rush to advance as fast as possible, you'll miss all sorts of fun things - like these hints!",
+    "Mixing certain liquids can have unexpected results - do things right, and you can make more of almost any liquid!",
```
bioboygamer, line 82: Don't know if this is technically true - not sure how up-to-date the wiki's liquid interactions page is. Still, thought it'd be important to let players know that mixing liquids can provide benefits.

objects/scienceoutpost/scienceinfobooth/scienceinfobooth.object
```diff
@@ -0,0 +1,91 @@
...
+	"Liquids will drain away if there's not a background block keeping them there - though oceans are an exception.",
```
bioboygamer, line 83: This is true of vanilla, but it took me an embarrassingly long time to realize it. With more space, I'd consider advising the player to use this to their advantage - either by adding background blocks to make raining liquids stick around for easy collection, or by removing background blocks to prevent flooding or drain pools of hazardous liquid.

objects/scienceoutpost/scienceinfobooth/scienceinfobooth.object
```diff
@@ -0,0 +1,91 @@
...
+	"Don't feel like using the teleporter to descend to the planet you're orbiting? Go down far enough outside your ship, and you'll fall to the planet below!",
+	"If some of my advice doesn't make sense to you right now, don't worry about it. You never know - it might come up later!",
+	"If you ever find alien ruins with grey bricks and purple or blue glyphs, you've probably found something built by the Ancients, a long-extinct race. Apparently, many Peglaci claim that there was another race even older than them, but not much is known about them.",
```
bioboygamer, line 86: Not sure how to best describe Ancient structures, but hopefully this should be sufficient. Not sure just how much about the Precursors is common knowledge - even the term "Precursors" as a specific race doesn't seem overly prevalent, so I figured I'd err on the side of caution.

objects/scienceoutpost/scienceinfobooth/scienceinfobooth.object
```diff
@@ -0,0 +1,91 @@
...
+	"Speaking of, I'm actually a Peglaci - I'm just using ultrasonic vibrations to create the illusion that I'm a Hylotl.",
+	"Why am I doing that? Simple - the person who added me is garbage at pixel art, and this sprite already existed."
```
bioboygamer, line 88: This is mostly a joke on the fact that the sprite used is exactly the same as the info terminal at the beginning of the game. It's true that I have zero talent when it comes to pixel art, or art in general, so I'm certainly not the person to draw up new sprites for something like this. Also, thought it'd be neat to reference the Peglaci's sound-related abilities, which I believe are otherwise unrepresented so far?

sayterdarkwynd:
these look good. Thanks man. I'll get them merged in likely for next release so I can rework the outpost and slot him in where he has a space rather than cramming.

PR #2998, opened 2021-03-31
Fix misc. typos

edwardspec:


sayterdarkwynd:
youll need to make a slight edit, as the peacekeeper bed was relocated

PR #3032, opened 2021-04-21
Updated crew nofall

Zarra-0:
- Fixed a bug relating to crew dying upon teleporting to the player.


Kherae:
move the status effect to the deprecated folder, and change the script as is done with the others in said folder.

PR #3088, opened 2021-06-04
Add Squamaeft race

Tigrex-Dai:
Add support for Squamaeft race

### DIFF
--- a/species/squamaeft.raceeffect
+++ b/species/squamaeft.raceeffect
@@ -1,0 +1,82 @@
+{
+	"stats" : [
+		{ "stat" : "maxHealth", "effectiveMultiplier" : 1.05 },
+		{ "stat" : "protection", "effectiveMultiplier" : 0.9 },
+		{ "stat" : "fireResistance", "amount" : -0.45 },
+		{ "stat" : "poisonResistance", "amount" : 0.4 },
+		{ "stat" : "cosmicResistance", "amount" : 0.1 },
+		{ "stat" : "radioactiveResistance", "amount" : 0.1 },
+		{ "stat" : "poisonStatusImmunity", "amount" : 1 },
+		{ "stat": "jungleslowImmunity", "amount": 1 },
+		{ "stat": "fumudslowImmunity", "amount": 1 },
+		{ "stat": "mudslowImmunity", "amount": 1 },
+		{ "stat": "fuclayslowImmunity", "amount": 1 },
+		{ "stat": "blacktarslowImmunity", "amount": 1 },
+		{ "stat": "slushslowImmunity", "amount": 1 }
+	],
+	"diet" : "omnivore",
+	"envEffects" : [
+		{
+			"biomes" : [ "jungle", "thickjungle", "ocean", "oceanfloor", "toxic", "toxicoceanfloor" ],
+			"stats" : [
+				{ "stat": "powerMultiplier", "effectiveMultiplier": 1.1 }
+			],
+			"controlModifiers" : {
+				"speedModifier" : 1.1
+			}
+		},
+		{
+			"biomes" : "hot",
+			"stats" : [
+				{ "stat" : "maxHealth", "effectiveMultiplier" : 0.9 },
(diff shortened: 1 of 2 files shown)


sayterdarkwynd:
all in all, reasonably balanced even considering the new method we are using to determine racial stats.

However, if you want "proper" balance on the new sytem i recommend the following:

1. remove both Spear/Shortspear and Bow bonuses from the list
2. replace both with spearMastery 0.15 and bowMastery 0.15
3. leave the Harpoon as-is

that should bring them roughly into line with the rest of the racial rebalances.

I'd also suggest a rebalance on those resistances. Nothing surpasses 30 now save for very rare exceptions (the key is in them being detrimental but not enough to make certain biomes nearly impossible)

Tigrex-Dai:
> all in all, reasonably balanced even considering the new method we are using to determine racial stats.
> 
> However, if you want "proper" balance on the new sytem i recommend the following:
> 
> 1. remove both Spear/Shortspear and Bow bonuses from the list
> 2. replace both with spearMastery 0.15 and bowMastery 0.15
> 3. leave the Harpoon as-is
> 
> that should bring them roughly into line with the rest of the racial rebalances.
> 
> I'd also suggest a rebalance on those resistances. Nothing surpasses 30 now save for very rare exceptions (the key is in them being detrimental but not enough to make certain biomes nearly impossible)

Can you tell me where I can see the details about the 'bow/spearMastery' ? and we'd like to keep the resistance thing though.

sayterdarkwynd:
Masteries
-----------------------
Mastery is basically replacing all the old weapon bonuses

it improves the weapons across the board in various ways (depending on the weapon). In the case of bows: damage, draw speed, and all those other things can be affected, rather than just a single value or two that you give to the race.

Masteries allow us to do all sorts of funky things with them, such as provide *different* bonuses based on scenarios (two handed, one handed, dual wielding, with shield, without shield, when running, certain combo hits etc etc.

they are applied at the top of the raceeffect file , and are listed as   "weaponnameMastery" (bowMastery, rapierMastery, assaultrifleMastery, etc)

You can see more specifics in various weapon type lua files. the next release will be relocating them and adjusting how they apply  and moves all of the data to new centralized locations for ease of use.

Racial Stuff
----------------
You can keep the penalties and bonuses if you like, but you're shooting yourself in the foot on hot worlds. There's a reason we no longer use resistances that low in the negatives: after years of playtesting its one of the -least- liked parts of gameplay.  By clamping it to around -30 you ensure that you are penalized but that it won't impact late-tier play in such a way as to cause undo stress and frustration. If you reduce that, though, you'll also need to reduce something else by about 10 points to be balanced again (penalties add points to spend on bonuses)

Calculating all your current bonuses, you come to just shy of the "100" points each race is limited to. So, by leaving it as-is you'll be within the bounds of that balanced state. Ultimately your call, but I fundamentally disagree with resistances that high/low anymore except in rare cases. (also...why poison? Amphibians are notoriously sensitive to such things because of their skin.)

I think it would be **_far_** cooler to specialize this race a bit more to really sell the amphibian aspect of them. They are a unique addition to the games modded races and deserve to feel truly special!

Tigrex-Dai:
I'll look into the weapon Mastery ;)
About the racial stuff, Maple(mod creator) said: (I've translated it)
Squamaeft's lore is that they were born in an extreme environment, so they have strong resistance against radiation/poison. However, because of amphibian, which is the very reason, they are afraid of fire.
But one thing you've mentioned makes sense that in Starbound fire planets are often in late-game, so too low in fire-resistance maybe a nightmare to players. Though Avali......
You can find detailed description of Squamaeft race by looking into our Steam Workshop page: They have poison even in the body of themselves, this refers to the salamander irl. Because they are also amphibian so water is necessary for them, though they won't be affected by poisonous water, dry/hot environment is fatal to them. I know that nowadays new races with just simple stats changed are many, but I think knowing the design of the race in advance is better.

Tigrex-Dai:
Maple: poison+35% fire-30% radiation+15% will do?

sayterdarkwynd:
Thats more reasonable for players. Ultimately its up to you guys.

PR #3089, opened 2021-06-05
Sending shipfiles from reddit convo

TychaiosIlithios:
Was making ship for myself. Combination of being well enough made to just not repeat tedious tasks when switching over and over, at the same time since my original intent was to craft planetside, the hitech on ship was mostly for decoration. Did a hasty nerf in file 2 as an example. 

Now check it out - if it is acceptable, good - anyhow gonna make a few more, now more dedicated to role of proper starting ships.
[clean unnerfed.txt](https://github.com/sayterdarkwynd/FrackinUniverse/files/6603310/clean.unnerfed.txt)
[nerfed example.txt](https://github.com/sayterdarkwynd/FrackinUniverse/files/6603314/nerfed.example.txt)


sayterdarkwynd:
these text files are of no use to us at all.

I had assumed you would be submitting Tiled files, which is what is required for FU BYOS ships to be integrated.

PR #3093, opened 2021-06-08
Disabled Bee Flower recipe

Zarra-0:
(Not changelog) For wiki purposes. Recipe was unlearnable


sayterdarkwynd:
Then it should be learnable, not disabled.

PR #3125, opened 2021-07-12
Feed-Munity description

hugojmaia:
Changed "Immune to Hunger Penalties" into "Immune to Diet Restrictions" to better match what the augment does.

Fix for when the player has 100% fire res but not lava immunity, resulting in rapid damage ticks.


sayterdarkwynd:
A big solid NO to fire resistance giving lava immunity.

If you want lava immunity, get a lava immunity effect.

giantcavespider:
100% fire resistance already give lava immunity, just in a borked way where you get repeated attempts at fire damage that do nothing, because you have 100% resistance

PR #3133, opened 2021-07-20
Some revisions, fixes, and additions

Kherae:
TL;DR:
fixed peblit having -2500% fire resistance, fluctuating ship mass and scroll of devastation aug slot. timewinderschnapps added but not implemented, unknown world hazards revised and expanded including inclusion of strange seas in the presence of 'revert' mods and additional time based effects. added support for arcana's arcanians. revised energy shell to have caps. Warning: Unknown worlds will reroll their status effects.

Long version:

Added Arcana to includes. Added support to tricorder for the Arcanian race. Added patch to display their stats, link to the relevant file is included in the patch, as the raceeffect file is in arcana itself.

Due to seeing the racial tech of the arcanians (+999 amount to energy regen. effectivemultiplier of 0 for regen block time), the following changes have been made to Energy Shell (wraith armor set bonus):
Block timer now has a minimum of half a second.
Regen is capped at 50% per second.

corrected Scroll of Devastation not having an augment slot.

fixed peblit having -2500% fire resistance.

fixed a bug with fluctuating ship mass due to a previous reorganization of the shipupgrades script

added handling for people using mods that add oceans to strange seas other than the default alien liquid. they will function like unknown worlds if that is the case.

temporal loop effect added: this records the player's position and resources (current values and lock state) at init, and at terminus will restore the saved values.

added 'timewinder schnapps' (timewinderschnapps) consumable. no recipes added, no pixel value set, no rarity set. Applies drunk, slow, and temporal loop for 5 seconds.

further adjustments to FU's time freeze effects to function more effectively on players.  This includes preventing most weapon use and disabling movement capabilities such as techs or grappling hooks. glitchtops now use one of FU's timefreeze effects instead of vanilla's. should largely be no difference.

implemented temporal instability: this 'weather' effect has a variable period from 5 to 30 seconds. there is a 5% chance to freeze the player in place for 0.5 to 3 seconds, preventing them from moving, attacking, or taking damage. otherwise, the other effects are rolled: a 17% chance to be caught in a temporal loop, a 51% chance to get one of two effects (fifty fifty that either will occur if this rolls): a slow, or a speed boost.

Revised handling on unknown world effect lists, including merging some lists and adding a wider range of effects
this includes some existing effects from FU, as well as the new effect 'temporal instability'. added a specific variant of insanity to replace the unresistable variant in unknown worlds only. this one requires 80% mental protection, otherwise cannot be mitigated.
furthermore, increased the amount of categories that will be rolled by approximately 50%.

BIG FAT WARNING,  PEOPLE ARE GONNA SCREAM: Version increased on Unknown world configs. previously rolled values WILL be rerolled. Your unknown world could be nicer, or it could be a nightmare. Enjoy.


thebestcodertoevercode:
So what's the point of trying to patch a patch? I don't see any point in this other than being pedantic. https://github.com/sayterdarkwynd/FrackinUniverse/pull/3133/commits/e61dedbfdc7c50dd453f0214c78f11305355bc7b

PR #3141, opened 2021-07-24
Minor fixes and Elduu weapon elements

hugojmaia:
Reloader augments now match rarity and price of their ammo counterparts.
Stack limits of some seeds and produce now match the rest.
Added elements to Eld'uukhar weapons.
Matched the tier of Eld'uukhar weapons to their crafting materials.


Zarra-0:
You've not changed the damageSourceKind parameter for the weapons to match their damage type, [see here](https://github.com/sayterdarkwynd/FrackinUniverse/blob/master/items/active/weapons/melee/broadsword/telebriumbroadsword.activeitem#L30) for an example. Disregard this if you've tested that it works.

The tier changes seem unnecessary when the elduu weapons already cover tiers 1-6/7 without gaps. Does each tier still have a 1h and 2h weapon? Changing the tiered metal required in the recipe is probably preferable to changing the weapons.

hugojmaia:
Apparently eld'uukhar weapons do not list the damage type in the special attack. I'll give it a test.
The tier thing is inconsistent, and does not match the materials the weapon is made out of, this is similar to when aetherium weapons were all changed to tier 7, as there were some that were t6 and some were t8.

sayterdarkwynd:
yea....tiers themselves don't need reworking imo, but your reasoning is sound enough so we'll try it.

PR #3146, opened 2021-07-26
merge upgrade interfaces into one

giantcavespider:
fu_multiupgrade and fuweaponupgrade have been replaced with a single interface+script to do the job of both

more testing may be helpful as there might have been things I've overlooked

some general user-facing differences:
upgrade table can now have target level selected
cosmic crucible can now upgrade tools
upgrade kits are treated as simply another currency, meaning it's easier to upgrade multiple levels with them
upgrade kits can no longer "downgrade" overleveled items, only a crucible can do so
some tools used to have their level set to 20 when fully upgraded, now the max level can vary, and attempting to "infuse" them at a crucible will result in downgrading their displayed level, with no other effect

also fix an edge case with the crystal skull code and fix the gel descriptions


Kherae:
added my own stuff to this as it was kinda needed:
increased button lockout time to 1.5 seconds, to give people more time to type out their numbers
revised the 'overloaded UI' text to be less mean and more informative.

PR #3156, opened 2021-08-06
Bounty Quest Unknown World Ban

rylasasin:
Bounty Quests are now banned from taking place on Unknown Worlds, as (especially on early tiers) it makes some bounties impossible to complete.


sayterdarkwynd:
Technically that was the goal: make bounties hard. But, a logical change in this case :)

PR #3177, opened 2021-09-21
Added "Leave One" and "Only Stack" functions to ITD

mousefrost:
Alongside the stack and slot split toggles in the item transference device, I added new "leave one" and "only stack" options as well. The former will always leave at least one item behind in every source slot/stack before transferring (it does this check early allowing it to be efficient and to take the 'minus one' item count into account before for item count calculations down the line). The latter only adds items to existing stacks in output slots, functionally treating empty, but valid slots as "full" for the purposes of round robin logic. 

Not too big a change, all in all its about a 50 line change between the gui config and lua scripting (the diff makes it look worse, wish LUA had a 'continue' function. If you can diff with whitespace ignored it makes it much clearer). Performance impact should be minimal, I tested both options together, as well as with the stack splitting toggles/filters mixed in. Everything seems to work correctly.

PlHqPVO

Using two ITDs and a storage item between them, this would effectively allow for "infinite" filtering options. Good example for this is a drop box that will automatically move siftable blocks and crushable blocks into their own separate containers that feed into the sifters/crushers. An ITD configured with "only stack" would move items from the drop box to the "siftables" box, provided you put any new block you encounter that is siftable into the sift box one time. Another ITD configured with "leave one" would then take these siftable blocks (but always leave one behind, so that the first ITD can still move blocks into it from the drop box), and send them to the sifters.

### DIFF
--- a/interface/kheAA/kheAA_router/kheAA_routerGui.config
+++ b/interface/kheAA/kheAA_router/kheAA_routerGui.config
@@ -528,6 +528,28 @@
 			"base": "/interface/kheAA/kheAA_router/slotItemGrey.png",
 			"baseImage": "/interface/kheAA/kheAA_router/slotItemGrey.png",
 			"baseImageChecked": "/interface/kheAA/kheAA_router/slotItemGreyChecked.png"
+		},
+		"leaveOneItemMode": {
+			"callback": "leaveOneItemToggle",
+			"type": "button",
+			"checkable": true,
+			"caption": "Leave One",
+			"hAnchor": "left",
+			"position": [129, 17],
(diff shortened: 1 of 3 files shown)


Kherae:
you tested absolutely every configuration possible to make sure all ITD functions still work properly, correct?

mousefrost:
Yeah, it works as expected with all the permutations of the even split and slot split functions. With the options turned off, the ITD should function exactly as it does now. I tried to make the changes as non-intrusive as possible. Filters should not be affected, but I lightly tested those as well too.

Kherae:
as long as you tested to make sure nothing broke, i'm fine with this stuff. had zero plans to implement them myself, however, as I felt it unnecessary.

PR #3192, opened 2021-10-15
replaced disgusting implementation of madness bombs

Kherae:
removed bmain.lua override
changed teamswap status effect to a charm effect, converting enemies to the opposite side
added teamshuffle effect which randomly changes a target's side periodically, while active.
teamshuffle and teamswap now have vfx, last 15 seconds, and restore the target's side upon expiry.
madness bomb throwable uses teamshuffle, instead of teamswap.
added charm bomb item, which does no damage and applies teamswap.  no recipe implemented yet.


sayterdarkwynd:
Please stop implementing things that replace existing functionality and then not include a recipe for them.

PR #3193, opened 2021-10-15
Remove Hardcoded limit making bees not work on ships.

V2LenKagamine:
Simple change to bees removing the hardcoded "don't work on ships" setting the "unknown" Biome to 0 [Deadly] For all bees

Yes, this would allow a modder to make a patch allowing bees to work on ships at half or full efficiency.

Why this should be added: Hardcoded bad, moddability good.


sayterdarkwynd:
Bees are not intended to work on ships. Period.

PR #3215, opened 2021-10-31
Bugfix: Replace unnecessary category 'upgradeable epp'

ulrellik:
This dedicated category gives you the impression there's smth special about this EPP regarding upgrades, but there's none ... It can be upgraded like other EPPs too, that don't have any special category. Even the batterypack_morphite that's part of this kind of EPPs doesn't have category 'upgradeable epp'.

This PR removes 'upgradeable epp' and replaces it with vanilla EPP category. I had no crashes/problem after this change during my tests and the EPP worked as expected.

### DIFF
--- a/items/armors/backitems/erithianalgaepack/batterypack_erithian.back
+++ b/items/armors/backitems/erithianalgaepack/batterypack_erithian.back
@@ -7,7 +7,7 @@
   "description" : "This lovely piece of gear grants: Energy x1.1, +10% E. Regen, -5% E. Block, protection from electrical storms and a slight ^#00aa00;glow^reset;.",
   "shortdescription" : "Erithian Energy Pack",
   "tooltipKind" : "baseaugment",
-  "category" : "^#e43774;Upgradeable EPP^reset;",
+  "category" : "enviroProtectionPack",
   "acceptsAugmentType" : "back",
   "radioMessagesOnPickup" : [ "pickupepp" ],
   "maxStack" : 1,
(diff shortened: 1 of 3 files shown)


sayterdarkwynd:
its not a bug, nor a fix. It's simply a change :)

Kherae:
I believe the category is for EPPs that are used as materials to make other EPPs?

ulrellik:
@sayterdarkwynd haha yeah it depends on the viewpoint, to be coherent 'batterypack_morphite' should also be of category 'upgradeable epp', since it's used for building 'batterypack_quantum', but it isn't, so this fixes this incoherence, thus bugfix :-D

@Kherae Yes I suppose this might have been the idea, but why aren't all the other EPPs that are also used as materials for upgrades of the this category, like batterypack_morphite, radiationprotectionback, xenonpack etc.? When I first saw those EPPs, I tried to upgrade them with the Tricorder ... it was confusing for me, but that is my personal opinion. You guys are the experts, I won't be offended if this PR is rejected, this is more a suggestion.

Kherae:
FU is full of inconsistencies like that, actually. personally i'm in favor of this PR. but if sayter wants to keep these as they are, then you can then do an about-face and go the other direction with the changes; make 'em all have the same category if they're used as a material for another EPP.

sayterdarkwynd:
nah, im good with this as well. Just havent gotten around to merging it yet. The category no longer serves a purpose in Research progression.

PR #3216, opened 2021-11-03
harden npc.lua against idiots

Kherae:
hardened /scripts/actions/npc.lua against improperly formatted status text due to idiot mods. it will now display a more meaningful error to the log.
 
 
 side note: maple32 is incompatible until they fix their patches to /npcs/statuses.config. this script changes fixes npcs despawning, but does not fix any broken recruits

### DIFF
--- a/scripts/actions/npc.lua
+++ b/scripts/actions/npc.lua
@@ -1,39 +1,43 @@
 require "/scripts/vec2.lua" -- added for npc aim fix attempt
 
 function randomizeStatusText(args, board)
-  local personality = personalityType()
-  local options = nil
-  if math.random() < 0.3 then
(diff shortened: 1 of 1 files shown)


ulrellik:
@Kherae Ohhhh I think this will fix the error I encountered:

> [16:58:18.786] [Error] Exception while invoking lua function 'update'. (LuaException) Error code 2, [string "/npcs/bmain.lua"]:162: (StarException) Lua Exception caught running action node randomizeStatusText in behavior crewmember: (LuaException) Error code 2, [string "/scripts/actions/npc.lua"]:11: (LuaConversionException) Error converting LuaValue to type 'N4Star5MaybeINS_6StringEEE'

It's not only Maple32 that causes this problem I believe.

PR #3217, opened 2021-11-04
Zerchesium Bar description

Zarra-0:
- Updated the description of Zerchesium Bars.


Hyperfire:
Forgive me if i'm wrong, but Zerchesium is a metal (Melted from ore), not an alloy, since a alloy is when combining 2 or more ingredients together, like bronze (Zinc + Copper). This is from a dictionary "alloy, metallic substance composed of two or more elements, as either a compound or a solution."

Pardon me for this as I dont mean any disrespect, its just going to be a typo later on. Just change alloy with metal.

PS: A alloy is also a metal, so it makes more sence.

sayterdarkwynd:
comments on git PRs are not the place for interrogative questions.

Zarra-0:
> Forgive me if i'm wrong, but Zerchesium is a metal, not an alloy, since a alloy is when combining 2 or more ingredients together, like bronze (Zinc + Copper). This is from a dictionary "alloy, metallic substance composed of two or more elements, as either a compound or a solution."
> 
> Pardon me for this as I dont mean any disrespect, its just going to be a typo later on. Just change alloy with metal.
> 
> PS: A alloy is also a metal, so it makes more sence.

Zerchesium is an invented material which isn't stated to be an element, defining it as an alloy doesn't contradict other lore

PR #3236, opened 2021-12-04
Reduced number of doll immunities

Zarra-0:
- Creepy Doll enemies are now heavily resistant to poison and ice damage, instead of immune. They remain immune to physical and shadow damage.

### DIFF
--- a/monsters/dungeon/doll/doll.monstertype
+++ b/monsters/dungeon/doll/doll.monstertype
@@ -216,14 +216,14 @@
         "powerMultiplier" : {
           "baseValue" : 1.0
         },
-        "shadowResistance" : { "baseValue" : 1.0 },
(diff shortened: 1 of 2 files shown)


sayterdarkwynd:
They are supposed to be immune to all but a few damage types. This is the intent of their design.

Zarra-0:
So 100% vs physical and shadow, and 80% vs poison, ice, rads and electric with a lower movement speed?
VS 100% vs physical and shadow, and 70% vs poison, ice, rads and electric, like I have it currently.

Sounds reasonable. I changed them from immunity to high resistance because I didn't think it was reasonable for a player to be expected to have five damage types to reasonably fight the vanilla mobs they come across, especially when these can spawn on T2 worlds where the player doesn't have access to most of the game's damage types

PR #3266, opened 2022-01-10
Geodes, sneodes and armeodes

Zarra-0:
- Removed FU ores from the geode manually-opened treasurepool, increasing the droprate of samples.
- Horizontally aligned the heads of Peglaci Snowpeople with their other parts.
- Armour sets which buff mining laser damage now specify that they only buff combat damage and not mining damage.

### DIFF
--- a/asteroids_worlds.config.patch
+++ b/asteroids_worlds.config.patch
@@ -1,10 +1,7 @@
 [
-{"op":"replace",
- "path":"/threatRange",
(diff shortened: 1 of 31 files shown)


sayterdarkwynd:
sorry, but why are we removing FU ores from a list I fully intended to reward FU ores from? I'm not against it but it seems to me this was done "just because".

PR #3270, opened 2022-01-17
armor conversion, set inheritance, and minor fixes

Kherae:
short version:
fixed wanderer having 0.02% crit chance instead of 2%
numerous armor system back-end changes
upgraded sets now contribute toward previous sets' bonuses, in case of incomplete upgrade.
sanguine set no longer grants health regen, and instead grants 0.5% life leech, doubled when wielding the appropriate weapons.

long version:
migrated remaining tier 1 to tier 3 armors to normal system so people wont lose their buffs when using mechs
fixed wanderer having 0.02% crit chance instead of 2%
finished implementation for that little bit of forgiveness for armor set bonuses I did with recon: when wearing 2 pieces of leviathan and 1 nautilus, the player will get nautilus, but not leviathan. etc. 
sanguine set no longer grants health regen, and instead grants 0.5% life leech, doubled when wielding the appropriate weapons.
fixed some back-end issues with tagcaching and setbonuses_common regarding two-handed state.
removed a ton of trailing spaces


sayterdarkwynd:
was the sanguine set change for any reason other than giving life leech to something else? I'm fine with it, just wondering what prompted the change :)

PR #3281, opened 2022-01-31
WIP: terminal: add request all items shown

Nephos:
Hello,

I've needed a button to collect all resources I filter when using item network on lots of containers (it does not always stack well).
Here my proposal to add such button "Request Shown" bottom right of the window.

Notice it's my first MR on a mod (ever) so be extra careful if you accept to merge :)
The button seems a bit off and maybe too much text too.

Screenshot_20220131_224316
Screenshot_20220131_224329

Cheers :love_you_gesture:

### DIFF
--- a/interface/kheAA/kheAA_terminal/kheAA_terminalGui.config
+++ b/interface/kheAA/kheAA_terminal/kheAA_terminalGui.config
@@ -105,6 +105,14 @@
 			"caption" : "Request All-1",
 			"position" : [130, 38]
 		},
+		"requestShown" : {
+			"type" : "button",
+			"base" : "/interface/kheAA/kheAA_terminal/buttonIO.png",
+			"hover" : "/interface/kheAA/kheAA_terminal/buttonIOH.png",
+			"press" : "/interface/kheAA/kheAA_terminal/buttonIOH.png",
+			"caption" : "Request Shown",
+ 			"position" : [130, 6]
+		},
 		"requestOne" : {
 			"type" : "button",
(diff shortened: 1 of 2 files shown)


sayterdarkwynd:
Not adding a button that the user cant read, certainly. Text needs to be more concise. Also awkward placed below all of the other "request" buttons. 

I'm also not really certain this is at all necessary. What purpose would this serve?

edwardspec:
Also needs some "are you sure?" type of safety against accidental clicks, especially when user didn't search first. Imagine accidentally retrieving 1000 stacks.

Nephos:
> Not adding a button that the user cant read, certainly. Text needs to be more concise. Also awkward placed below all of the other "request" buttons.
> 
> I'm also not really certain this is at all necessary. What purpose would this serve? I'm curious, as wiring is not my forte.

When wiring lots of containers you can have multiple stack on the same item (and sometimes a LOTS of stacks). It takes lots of space so I need to multiply the container or request each stack manually. It also happens when I've several full stack tot retrieve. However I can understand you skepticism!

I will try to rename it "Request All" maybe, it is shorter.

Agreed with the "are you sure", It's gonna take me some times since I know nothing about it :D

sayterdarkwynd:
Allright, give it some time and code it up and we'll see how it looks at that point. I'm not sure its something we need, but I am all for trying it out.

sayterdarkwynd:
I am assuming this ticket is dead if it has not yet been addressed?

Nephos:
Yes all my friends stopped playing it so did I.

PR #3297, opened 2022-03-12
NPC Boomerang!!

rylasasin:
Fenerox Guards sometimes have Wooden Boomerangs instead of Bows.

Hylotl Ninjas have Throwing Stars instead of that silly tungsten bow.


rylasasin:
going to readjust file structure

PR #3299, opened 2022-03-19
make colony core work even if unloaded

night-glider:
As I mentioned on discord, we made colony cores work even if unloaded.

The algorithm is simple:
1. When core initializes, it will check last time it was loaded
2. Then it will check current time and find out how many ticks were missed
3. Then it will compensate all missed tick on the next frame (simply adding ( average_rent*missed_ticks ) amount of pixels)

Co-Authored-By: Kherae <25236459+Kherae@users.noreply.github.com>


scripts/colonycorefu.lua
```diff
@@ -3,76 +3,56 @@ require "/scripts/kheAA/transferUtil.lua"
...
 	wellInit()
 	setDesc()
+
+	--figuring out how many ticks we missed
+	local leftTime = config.getParameter("leftTime")
+	if leftTime then
+		local diff=os.time() - leftTime
+		if diff>0 then
+			diff=math.min(math.floor(diff/rentTime),math.floor(maxOfflineSeconds/rentTime))
+			offlineTicks=diff-1
```
edwardspec, line 30: Need a sanity check that `offlineTicks` is not negative, e.g. `math.max(0, diff - 1)`

night-glider: diff can not be negative since we already check if it's greater than zero. And diff can not be float since both os.time() and leftTime are in POSIX format (integer amount of second)

edwardspec: You are setting `diff` to a different value after the `diff>0` check, and it can become 0 there.

night-glider: OK i'll add additional check tomorrow, just in case

night-glider:
ok now we're safe to go

scripts/colonycorefu.lua
edwardspec: `offlineTicks=math.max(0, diff - 1)`

scripts/colonycorefu.lua
```diff
@@ -3,76 +3,57 @@ require "/scripts/kheAA/transferUtil.lua"
...
+			math.max(0, diff - 1)
```
edwardspec, line 31: This does nothing. Should be above:
`offlineTicks=math.max(0, diff - 1)`

edwardspec: Make sure to test (in-game) that everything works as expected:
1) when returning to the planet with Colony Core after 4-5- minutes have passed (which is more than `rentTime` seconds)
2) when returning to the planet immediately (before `rentTime` seconds: beam up, then return to the planet)

night-glider: > This does nothing. Should be above: `offlineTicks=math.max(0, diff - 1)`

That's why you should never code immediately after sleeping for 6 hours. Fixed this nonsense.

night-glider: Ok this additional check breaks our system. It causes our colony core to give you 1tick worth money EVERY time you beam on the planet (even if only 1 second passed). It happens because offlineTicks turns 0 and the next frame we add `(1+offlineTicks)`  ticks. Since offlineTicks=0, we add at least 1 tick every time we load our colony core.

So.. offline ticks should be -1 if we didn't wait enough time.

so the final check should be like this `offlineTicks=math.max(-1, diff - 1)`

night-glider:
> Make sure to test (in-game) that everything works as expected:
> 
> 1. when returning to the planet with Colony Core after not being on that planet for 4-5 minutes (which is more than `rentTime` seconds)
> 2. when returning to the planet immediately (before `rentTime` seconds: beam up, then return to the planet)

Tested a couple of times. It works as expected in both cases.
1st case: gives all the missed money.
2st case: does not give you anything

There is only 1 bug. It does not work on the ship for some reason.
It just does not save its `leftTime` in its config. Dunno why this happens.

sayterdarkwynd:
in the case of the ship, what is the fallback then? Default behavior, I imagine?

night-glider:
> in the case of the ship, what is the fallback then? Default behavior, I imagine?

Yeah, it just does not compensate missed ticks. So no errors and crashes.

sayterdarkwynd:
cool. I imagine that's probably good to go then if the other guys similarly have no objections.

sayterdarkwynd:
When you get a moment please clean up those trailing whitespaces so the linter stops freaking out and then I think we're good.

night-glider:
> When you get a moment please clean up those trailing whitespaces so the linter stops freaking out and then I think we're good.

did it.

PR #3301, opened 2022-03-24
NPC Throwing Items

rylasasin:
A few throwing items for NPCs. 

Unfortunately, the throwing spear, throwing kunai,  and crystallite shard have an issue where the projectile doesn't flip like it should. I can't figure it out and need help with it. I've left these items out of the NPC's weapons list for now. But if someone could get those three working I'll add them to the fenerox guards and pavilian archer (ninjas) respectively.

Currently, Fenerox Guards use wooden boomerangs sometimes instead of bows. Ninjas no longer use bows, but now use a variety of throwing items. Most of the time ninja stars, but sometimes weapon-plants (ignus, shock shroom, deathblosom, etc)


sayterdarkwynd:
looks like you need to merge with master to update your recipes

PR #3304, opened 2022-03-27
BYO monster gravity

bum8844:
BYO monster gravity


sayterdarkwynd:
Not sure why this is necessary. Nothing spawns to fight in shipworlds.

PR #3305, opened 2022-03-28
use mech in byo ship

bum8844:
use mech in byo ship


sayterdarkwynd:
I suppose its not really a problem, but I am unsure exactly how or why this would be necessary. How do you even launch a mech in your own shipworld?

PR #3308, opened 2022-03-28
Bunnine Race addition

AkariEnderwolf:
I tried to do what I needed to to get the bunnine race added. Bunnine race not on workshop yet since I was trying to test compatability before uploading to workshop, and get a few more things ready for it.


AkariEnderwolf:
the internal name should be "bunnine"

sayterdarkwynd:
These are extremely OP.

AkariEnderwolf:
so, everything is working now and is hopefully properly balanced.

AkariEnderwolf:
what does the mastery system actually do?
Also, why have a point cost for biome immunity if it's not allowed? genuine question.

As for liquid nitrogen, couldn't find a thing for just the normal, attack inflicted, freeze status effect. The race normally lives on cold planets in lore, so they would resist or be immune to being frozen by enemy attacks.

AkariEnderwolf:
was trying to help with the metadata includes thing, and couldn't figure out how to undo a commit when it said there were conflicts

AkariEnderwolf:
apparently the conflict was the file ended with } when it should end with } ... no idea how that's a conflict.

PR #3322, opened 2022-04-17
Typo fix for the Vegan Cakes

hugojmaia:
All 4 vegan cakes had "/n" showing in their descriptions. The original intent was probably a new line "\n" instead.


sayterdarkwynd:
i believe Zarra already corrected this! Thanks though!

PR #3351, opened 2022-06-08
Proposed buff to alternator generator

anoobindisguise:
Alternator generator is currently the "tutorial" generator. It makes sense that it's pretty lackluster, you essentially use it to learn about how power works and then immediately ditch it when you have steam/solar/combustion. However it's so aggressively weak that it almost feels like a waste of resources to feed it fuel so I wanted to remedy that somewhat. 

To this end it still only provides 4w power, but the duration of its fuels have been increased to double that of a combustion generator's. In other words, it provides 1/3 the power of combustion for 2x the time giving it a total fuel efficiency compared to combustion of 2/3 the total j-output of combustion. However it's also still 1/3 the wattage as usual meaning combustion is the clear winner on both fronts, but is no longer such a massive improvement that it has both better durations of fuels on top of 3x the watts. 

I gave the cat fuel a nearly infinite duration mostly as a joke, as no one would be insane enough to waste the cat fuel on an alternator but if for whatever reason you do it will power it for about 40 hours.

This is experimental and I don't necessarily expect it to get merged. It's perfectly acceptable that the alternator is the worst generator on every front as it shows you a clear path of improvement for each type of generator you research. This generator provides the same power but for free (solar). This generator uses the same fuels but provides 3x the power or more (combustion). This generator is a weird sidegrade that has good power output and uses potentially more common fuels but takes up a ton of space and resources (steam). 

Additionally there was once a plan to make the alternator the "solid" fuel burner and the combustion engine the "liquid" fuel burner giving each one a distinct profile. This is another option but is more of a drastic change. As it stands I mainly intended to make putting fuel in an alternator feel like less of a waste than it currently does in my highly subjective experience.

### DIFF
--- a/objects/power/fu_alternatorgenerator/fu_alternatorgenerator.object
+++ b/objects/power/fu_alternatorgenerator/fu_alternatorgenerator.object
@@ -44,15 +44,15 @@
 
   "powertype" : "output",
   "acceptablefuel" : {
-	  "coalore" : 7,
-	  "logblock" : 7,
-	  "energywood" : 14,
-	  "goldenwood" : 10,
-	  "schrodingerscat" : 140,
-    "alienweirdwood" : 20,
-    "astralwood" : 30,
-    "redwoodbark" : 10,
-    "penumbralbark" : 11
+	  "coalore" : 16,
+	  "logblock" : 10,
+	  "energywood" : 50,
+	  "goldenwood" : 20,
+	  "schrodingerscat" : 120000,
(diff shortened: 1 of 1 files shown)


sayterdarkwynd:
Not sure on this one either. The idea is decent, and the reasoning is sound enough...I am simply unsure if this is in any way necessary. It's designed to be obsolete quickly, though this does mitigate it and extend its usefulness. I'll give it some thought!

PR #3352, opened 2022-06-08
craft swamp water

anoobindisguise:
Allows you to craft swamp water, unlocked at basic liquid crafting. Nothing too special.
It may be better to leave this liquid as a biome exclusive resource (assuming you want to centrifuge it for an infinite quantity of mud for some reason) however I am using salt water as a precedent for its inclusion as its circumstances and creation are similar to that of salt water.
I need to test this in game just to make sure it's fine first. I'm pretty sure I have the ids correct here but it doesn't hurt to be on the safe side.


sayterdarkwynd:
I....don't really see the point of this?

PR #3356, opened 2022-06-14
Reduce lava efficiency in dynamo

anoobindisguise:
Lava dynamos are a little too good. They're an excellent tutorial on power automation (hook up a well and a pump to the core and start producing cheap power) but lava was good enough that the other fuels weren't good enough of a step up IMO. It's still a clear improvement over the alternator but it should also incentivize producing more complex fuels (such as core fragments, volatile powder, or liquid iron, which can be refined from centrifigued lava as an example). 15w was just a little bit too high for the ease of automating it had and 12w is a little more fair.

It could even go down to 3w per lava but this may be too extreme.


sayterdarkwynd:
In your PR messages please include a point-form description of what you actually adjusted so its visible at a glance. Approving this one is a no-brainer, though.

PR #3372, opened 2022-07-07
Radioactive Rods in Manufacturing Station

Maxouille64:
These rods have a recipe in the Bio-Chem lab which has nothing to do with smelting, so it would make sense for them to be able to use that recipe in a manufacturing station. If it causes the smelting recipe to also become available at the manufacturing station, it's a harmless side effect for an overall improvement. There are many practical applications to this recipe being available to be automated, especially in the production of more powerful radioactive materials.


sayterdarkwynd:
Hard no. The manufacturing station does not make *any* bars or similar ores for a reason.

PR #3389, opened 2022-08-06
Candles

N1ffe:
Adds 2 new variants for beeswax candles (blue, flower, ghost, red and wax) - with two and three candles in single object. Got inspiration from a random indie horror game, and i think these would fit pretty well in cultist-themed buildings. I also did some adjustments to these existing candles:
- No longer burn underwater.
- Added missing novakid description for some.
- Adjusted glow colors and brightness so they're no longer incredibly bright.
- Adjusted light flicker to be the same as for vanilla candle.
- Simplified orientations config and removed ability to place them on walls.
- Made them interactive and added some interaction sounds.
- Slightly adjusted inventory icons for blue, ghost and red candles.

Imgur link (i use dynamic lighting on them but you'll still be able to compare them to vanilla torch (also red and flower candles not shown as they have the same glow colors as wax candle)): https://imgur.com/a/XrdUXzl


sayterdarkwynd:
did you add recipes and research entries for them or are they purely deco? if so, we should add them to a drop list somewhere so players can obtain them on their own. Or, ensure they are scannable for the pixel printers

PR #3406, opened 2022-08-29
Mite warning system fixed.

Davee47:
Mite warning system fixed. Apiary will display mite symbol if mites are present.
The thresholds for rotten/infested apiary looks have been reduced as well.


Davee47:
Also queens will no longer "magically" disappear. Code is adjusted so that mites do damage to queens only if 1 mite is present. Previously, mites killed the queen if there was mites greater than 0 but less than 1

PR #3407, opened 2022-09-02
Update beeData.config

Davee47:
Fixed some bee subtypes to not be strictly inferior to their comparable subtype. Made them equal in output to the superior subtype instead.
Goldensaint, oremason, inventor, reaper and red-banded all had strictly inferior subtypes.


Zarra-0:
goldensaint and purewing have different goldensand output timers

Davee47:
Thank you! Got that fixed

PR #3412, opened 2022-09-19
Change elduukhar flight icon and remove feather particles

VladislavRazoronov:
This code changes Ed'uukhar flight icon to something more fitting and removes feather particles from ready effect.


sayterdarkwynd:
The feather particles are there for a reason. The effect must have particles to display that it is active, and none when it is inactive.

PR #3439, opened 2022-12-13
Make liquid irradium usable in fission reactor

anoobindisguise:
Irradium in its raw farm is a radioactive material that makes sense for a fission reactor to utilize. Fission reactors lost an important fuel in Hydrogen and is somewhat hard to use due to fuel scarcity so I believe an addition of a generally renewable (with steps involved) fuel would be good for its balance. 

Its specific numbers are up in the air. I have it at 12 between Tritium and Deuterium, while it is more common than Tritium or Deuterium and easier to produce, Tritium and Deuterium also aren't actually fissile fuels and there's a decent argument to just remove them entirely and replace them with Irradium. Thoughts on this are appreciated.

### DIFF
--- a/objects/power/isn_fissionreactornew/isn_fissionreactornew.object
+++ b/objects/power/isn_fissionreactornew/isn_fissionreactornew.object
@@ -51,6 +51,7 @@
 
   "bonusWasteChance" : 50,
   "fuels" : {
+      "liquidirradium" :      { "power" : 12,  "decayRate" : 30  },
       "tritium" :             { "power" : 10,  "decayRate" : 60  },
       "deuterium" :           { "power" : 15,  "decayRate" : 50  },
       "uraniumrod" :          { "power" : 20,  "decayRate" : 90  },


sayterdarkwynd:
Not sure this is in any way necessary, given how easy fusion/fission is to fuel in the first place. It's already among the best power sources you can produce! I don't necessarily object to it, and the balance seems okay...I just don't see the point. 

Sell me on it.

anoobindisguise:
My thinking is that as of present, the primary means of fueling a fission reactor is deuterium rods, as they do not produce enough tritium and toxic waste to self-sustain although they can increase their power output by diverting their outputs back into themselves. Their "primary" fuel types (uranium, plutonium, and other isotopes) are primarily obtained through active methods - mining radioactive planets specifically, but deuterium is the only really "renewable" way to fuel them. It takes 12 fusion reactors running off of hydrogen to make enough toxic waste to fully fuel one fission reactor with uranium rods (less if you are making enriched uranium but that is an extra-complex process) but if you have 12 fusion reactors you don't really need a single extra fission reactor. 
So irradium provides a means of supplying mass quantities of a low-quality fuel to fission reactors. While deuterium can be automated by mass centrifugation of dark water (or centrifuging 100x as much regular (sea)water), irradium provides additional means of fueling fission reactors by either being pumped from a rare core in irradiated planets or chained through liquid mixers (water->organic soup ->irradium or pus->irradium) providing additional avenues for fuel production and additional desirable planets for basing.
Another alternative is to allow poison to produce toxic waste, or allow Toxic planets' oceans specifically to produce toxic waste (replace the liquid with "toxic waste" which would present both poison and radioactive dangers, or replace with bio-ooze (and allow bio-ooze to produce toxic waste).  The main problem right now I see with Fission is that they are almost entirely reliant on non-automatable fuel sources and don't offer much of a benefit over, for example, placing 50 wind turbines which are easier to get and don't emit radiation. Fusion is more attractive due to ease of using helium-3 or hydrogen, and thus it's reasonable to just skip fission entirely since it's too much effort for too little benefit.

sayterdarkwynd:
I suppose. Not really sure why that needed a novel to describe though, heh.

I still don't see this as any sort of issue though: Frankly, I get annoyed at the constant badgering by players to change power to be "better" when there are a dozen types that serve a niche just so they can AFK more efficiently, which goes against the intent of the mod in the first place.

I'll approve this, but nothing more regarding power.

PR #3440, opened 2022-12-13
Add plutonium spawns to early planets

anoobindisguise:
Adds plutonium spawns to tiers 1-2 planets underground, mirroring neptunium distribution frequencies. This should make plutonium a little less rare because while it is a fairly low tier ore it only really shows up with frequency on higher tier radioactive planets.


sayterdarkwynd:
A hard no to this. It is not a t1-2 material and is not intended to be found that early.

PR #3444, opened 2022-12-21
Chewable magnet lasts 10x duration; recipe produces 1/10th the magnets

anoobindisguise:
The recipe used to make 10 magnets, but each magnet lasted a fairly short time. I wanted to reduce the total micro involved in using the magnets (having to constantly refresh the buff is annoying) and this accomplishes that without changing the balance of chewable magnets' ingredients.


sayterdarkwynd:
Simple and effective. approved.

PR #3451, opened 2023-01-16
Added full list of biomes to GPS

Kukagres:
Added full list of biomes to GPS. Some biomes are hidden from the list. Instead, you will get a red text saying that some anomalies were found, but you don't know in which layer.
In addition, each layer says in which position it begins and in which it ends, being able to locate you easily in them.


sayterdarkwynd:
Seeing some errors in the linter here. Fix those up and we're good to test.

PR #3464, opened 2023-04-01
Add files via upload

Adnarim7:
penumbra2 parallax fix removing floating pixels


edwardspec:
Can you feed these images to some PNG optimization tool? (their filesize will be reduced by more than 50%)

PR #3471, opened 2023-04-13
Add files via upload

Phosphra:
need new lunariswoosh sprites


Zarra-0:
Please don't change the rarity of crafted weapons/armours, their rarity reflects their base tier to make sorting through the Armoury easier.

I'm not a fan of the proposed Lunari Broadsword sprite, the current one has much cleaner lines and better fits Starbound's colour aesthetic. Why reduce the size of the sword?

sayterdarkwynd:
I concur. While the "new" sprite looks shinier, it is a departure from the SB style, which means it wont be approved in this case. More colors is not always better:)

Note that this is not saying your sprite looks bad. It doesn't. It simply doesn't fit SBs 4-shade style.

sayterdarkwynd:
Guess you have no intention of fixing this, so I am closing it.

PR #3498, opened 2023-08-21
Slopped variants of Copper and Silver

RockCat89:
Nothing more to say, adds "silversloppedpanel" as silver variant of slopped block and "coppersloppedpanel" as copper variant
both integrated into research tree and occupying tileID's of 6481 and 6482.

I have no clue what to do with files inside of tiled folder so I just added icons

It's been tested and works, tiles look good (Although I personally feel copper is a bit dim, but can't do much more with this palette)


Zarra-0:
You need to increment the version number at the bottom of the craftsmanship tree so that players who've already unlocked the node will get the new recipes

RockCat89:
> 

Yep, sorry didn't realize that, should be good to go now =)

PR #3500, opened 2023-09-08
Make taurikin omnivores

BitBlitObviMormon:
FU lists taurikin as herbivores when they are actually omnivores (fox taurs). This proposed change fixes that.


sayterdarkwynd:
Im leaning more toward leaving them as herbivores strictly because there are so few in the mod currently.

BitBlitObviMormon:
So you'd confuse players and screw with the lore because there's not enough herbivores??

PR #3518, opened 2023-12-09
Adding Mechineki FU Support

vlekje513:
The original Mechineki FU Patch was removed by the author, since then it has been re-uploaded by someone who happened to have the files. I wish to have the FU Patch built-in FU with my own balance of the race. I have played the game for I think over 100 hours with this race and would say it's largely balanced, it's good at some things, and bad at others, enough to make the experience unique and fun. It should also be balanced in according to the [balance sheet](https://docs.google.com/spreadsheets/d/1P_QSYx6OjVtiFy7ssmIgjB4E1NISqyBzgS-yGs_z6Ao/edit?usp=sharing).
simple picture

This is my first pull request so my apologies in advance if I messed anything up, I tried my best to do this correctly anyways.
This PR includes just **3** things, the **.raceeffect**, the **.species**, and adding the **mechineki in the race list of the statwindow**.

The only thing I'm uncertain of is whether it'll cause issues is it being dependent on both mechineki, and neki. I'm not sure how FU will treat clients that don't have any of these mods installed. I hope it simply gets ignored and I assume so since the race list is quite full of races, but I'm not 100% certain. Either way, even if it's incorrect it's probably easier to fix this mistake rather than asking about it before even opening this pull request.


vlekje513:
After checking the image, it seems the Health debuff (-0.85x) is missing from the weakness list.

sayterdarkwynd:
https://docs.google.com/spreadsheets/d/1hhGqycPcYUmlcSsPH5dn-Yvv_Il0D2X6KMIU8wxHbuI/edit?usp=sharing

this doc contains the actual balance charts for the currently supported races. Use the same metric for your balancing to be sure they work out to be approximately 100 points. If not, they need to be adjusted to that 100-ish threshold.

vlekje513:
> https://docs.google.com/spreadsheets/d/1hhGqycPcYUmlcSsPH5dn-Yvv_Il0D2X6KMIU8wxHbuI/edit?usp=sharing
> 
> this doc contains the actual balance charts for the currently supported races. Use the same metric for your balancing to be sure they work out to be approximately 100 points. If not, they need to be adjusted to that 100-ish threshold. Then you'll know they are balanced like the rest!

The balance sheet in my original post has the same values as the actual one as it's a copy, so I think it should align. Some rulings were a bit confusing for me in there like Snow Immunity and Slush immunity being separate or not and the Bee immunity not being counted for any points for the Glitch race, same for the GlitchLiquidEffect apparently not taking any points either.

It should be approximately 100 points from my interpretation atleast. I think it's quite balanced either way, with the Fist and Magnorb Masteries for example only working in close-ish range with a -15% health race it should balance it out anyhow. If you disagree however let me know here or in the discord and I'll discuss it or change something outright.

sayterdarkwynd:
Cool, so long as you paid attention. That's more than most do. You'd be amazed at some of the attempts to give races stats lol.

PR #3536, opened 2024-02-12
Adding effects for astral constructs

WaifuAquamarine:
effects for custom species that I literally put out today. the astral constructs


WaifuAquamarine:
ultimately decided to make a patch

PR #3538, opened 2024-02-17
Update shortswordcombo2.weaponability

MaxDreyko:
slightly modified stances and changed idle rotations for smoother animation
increased comboSpeedFactor
actual strikes and damage unchanged
reduced flash color


sayterdarkwynd:
seeing some issues here. Might just require you pulling the new update into your fork. Looks like the cloud stuff that was just added, which definitely wasnt your add.

PR #3546, opened 2024-03-13
Fix species description for Mechineki by adding include

vlekje513:
The Mechineki .species file now has a description to have built-in support with Race Traits, but this breaks the FU mechineki.species.patch somehow (I thought patches always applied later regardless of load order). I think an include so it loads after the initial file is there should fix it anyways. Since FU support is built-in for Mechineki I don't expect circular dependencies to be an issue.


sayterdarkwynd:
i dont understand why this is needed if the mechineki mod already includes support?

vlekje513:
The support it added was not for FU, but [Race Traits](https://steamcommunity.com/workshop/filedetails/?id=2622273194).
The change added the `/charCreationTooltip/description` that was previously empty. Now that it's not empty, FU's gets overwritten instead, is what I'm guessing.

Either way, the `mechineki.species.patch` does not work anymore, as the race description in FU (without Racial Traits of course) does not show the FU description.

PR #3553, opened 2024-04-01
fixes

Kherae:
Corrected happiness underflow in colony core system. Additionally, negative happiness will now delay the timer.

Reverted unacceptable change to protogen species, +20% healing was buffed to +120% healing in another user's commit. This has been brought back down to +20%.


GabrielF-C:
Isn't happy always equal to 0 on line 55 ?

PR #3578, opened 2024-07-08
Fix header of Snarshing Table window (it was saying Esoteric Research)

edwardspec:


Blkandwhtgaming (approved):
Thanks for the fast fix!

PR #3586, opened 2024-07-31
Update Novakid Brands to use Proper Placeholder Color

Skulgan:
Mainly so that modded recolors for Novakids work.


sayterdarkwynd:
You just need to run those images through PNGGauntlet and your PR should no longer throw an error.

PR #3590, opened 2024-08-24
backend fixes and stuff

Kherae:
short notes:
fixed callistan food rush
backend changes for slows and AI stat tracking and pathing
fixed many nonfunctional slows
cat is tired.

long notes:
corrected food rush on callistan/calskies. tentatively added(?) calliote.

gave runboostdebuff effects vfx, shortened base duration from 30s to 5. added runboostdebuff to crystalline shard projectile to match its subprojectiles

extensive update to fu's statusutil script to add automatic handling of some AI specific things for status effects, and to move the spike sphere speed shutoff to a centralized location.
redid a lot of status effects to move to the above, including various tile effects and armor set bonuses.

updated stat list documentation: jumpmodifier is not a controlmodifier. it is a stat used purely for AI. for controlmods use airJumpModifier

replaced all erroneous usage of runModifier (which does nothing whatsoever) with speedModifier. similar with jumpmodifier and airjumpmodifier.

due to the above, people may be slowed or crippled where they weren't before. some of the values were adjusted to proper values, some may require feedback.


Partonetrain:
The Calliote species name is indeed "calliote". Thanks for you work!

PR #3599, opened 2024-10-20
Add rarities filter to Armory

atom-skillz:
Adds a way to filter armors and weapons by tier, since in FU the rarity directly linked to the tier of the item. This will help players pick a weapon/armor more easily.

### DIFF
--- a/interface/windowconfig/armory.config
+++ b/interface/windowconfig/armory.config
@@ -315,6 +315,38 @@
       "type" : "radioGroup",
       "toggleMode" : true,
       "buttons" : [
+        {
+          "position" : [8, 60],
+          "baseImage" : "/interface/crafting/sortcommon.png",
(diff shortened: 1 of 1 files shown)


sayterdarkwynd:
I remain unconvinced that adding 4 new buttons is an answer to anything other than cluttering the interface

atom-skillz:
> While the idea is great and I like where the implementation sits on the interface: clicking on them crashes things because you didnt actually set up groups to do this. The game has no idea what to load. It doesn't filter using that parameter by default.
> 
> This also means you clearly did not test the function to see if it operated: please do not submit pull requests without testing.

Do you have the error message? I am new at modding but I have been testing since it yesterday while playing and i haven't had any issues with the change. 
20241021_11h58m51s_grim
20241021_11h59m06s_grim

PR #3604, opened 2024-11-04
Fix table used as int

### LINKED ISSUE #3603: attempt to perform arithmetic on a table value
```lua
[23:21:54.943] [Error] Exception while calling script init: (LuaException) Error code 2, [string "/items/active/weapons/ranged/gunfire.lua"]:82: attempt to perform arithmetic on a table value
stack traceback:
	[C]: in metamethod 'mul'
	[string "/items/active/weapons/ranged/gunfire.lua"]:82: in method 'calcAmmo'
	[string "/items/active/weapons/ranged/gunfire.lua"]:33: in method 'init'
	[string "/items/active/weapons/weapon.lua"]:119: in method 'init'
	[string "/items/active/weapons/ranged/gun.lua"]:38: in function <[string "/items/active/weapons/ranged/gun.lua"]:6>
```

at https://github.com/sayterdarkwynd/FrackinUniverse/blob/master/items/active/weapons/ranged/gunfire.lua#L82

cause some FU guns not to aim and can't shoot. For example `swtjc_ewg_commonrevolver`

linonetwo:
fixes #3603

I think some items may have buggy value in config, while I find the config in this repo is good

https://github.com/sayterdarkwynd/FrackinUniverse/blob/bdada8a6a79f882791b2876b506aa26469394f4a/items/active/weapons/ranged/pistol/swtjc_ewg_commonrevolver.activeitem#L72

so maybe bug comes from some other FU compat mod, that is outdated. But have guard in the code is always a good option. At least this make gun shootable again.

### DIFF
--- a/items/active/weapons/ranged/gunfire.lua
+++ b/items/active/weapons/ranged/gunfire.lua
@@ -35,8 +35,13 @@ function GunFire:init()
 	--self.magazineAmount = math.min(config.getParameter("magazineAmount",defaultMag),self.magazineSize) -- current number of bullets in the magazine
 	--self.magazineAmount=config.getParameter("magazineAmount",defaultMag)
 	self.magazineAmount=config.getParameter("magazineAmount"..self.abilitySlot,defaultMag)
-	-- how long does reloading mag take? Ensure reloadTime config value is a number with tonumber fallback
-	self.reloadTime = math.max(0, (tonumber(config.getParameter("reloadTime", 1)) or 1) + (tonumber(status.stat("reloadTime")) or 0))
+	-- how long does reloading mag take?
+	local reloadTimeConfig = config.getParameter("reloadTime", 1)
+	if type(reloadTimeConfig) == "table" then
+		reloadTimeConfig = math.random(reloadTimeConfig[1], reloadTimeConfig[2]) - Ensure reloadTime config value is a number
+	end
+	self.reloadTime = math.max(0, (tonumber(reloadTimeConfig) or 1) + (tonumber(status.stat("reloadTime")) or 0))
+
 
 	if (self.isAmmoBased==1) then
 		self.timerRemoveAmmoBar = 0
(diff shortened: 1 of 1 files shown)


edwardspec:
Intended handling was to randomly choose a number between 5 and 8.
This is how it works when other parameters (such as damage) are set to an array of two numbers.

linonetwo:
Hi, do you mean this is the Intended behavior in FU, but gunfire.lua forget to adapt to this usage?

Since the array `[5, 8]` in config is indeed interpreted as a table in Lua, which caused the original error because arithmetic operations can’t be directly applied to tables.

But I guess, the array `[5, 8]` in config should not be use directly here, it should generate an item data with a int instead of array. And we get the item data instead of config here. I'm new to modding so I'm not sure here.

linonetwo:
I think this PR is correct, I just add a guard to parse the int, if other mod copy the array to instance instead of generate an int before copy, this will fix it.

edwardspec:
Currently this PR is incorrect, because while it fixes the Lua error, it replaces it with silent incorrect behavior:

```lua
tonumber(status.stat("magazineSize")) or 0
```

Situation: `"magazineSize" : [5,8],`
Expected behavior: `magazineSize` should be set to random value between 5 and 8.
Current behavior of your code: value of `magazineSize` is considered incorrect and set to default value of 0.

edwardspec:
Question: have you actually encountered a Lua error when running the game? Or are you theoretically assuming that it might happen?

Because I see correct handling of `magazineSize` being a table in the weapon's buildscript:

https://github.com/sayterdarkwynd/FrackinUniverse/blob/bdada8a6a79f882791b2876b506aa26469394f4a/items/buildscripts/buildweapon.lua#L272-L276
https://github.com/sayterdarkwynd/FrackinUniverse/blob/bdada8a6a79f882791b2876b506aa26469394f4a/items/buildscripts/buildweapon.lua#L290

linonetwo:
I understand, I update the code to handle randomlization at runtime, while it should have done in weapon creation time.

The solution at least make the weapon useable. While it may change after restart game

图片
图片

> Are you sure you don't have any incompatible mods that overwrite `items/buildscripts/buildweapon.lua`?

Seems it is impossible to know this in Starbound, I can't log who overwrites it.

But from the screenshot, looks like sometimes it has localization and sometimes not, maybe localization performs some bad patch.

linonetwo:
Some poor guys 5 years ago 

- https://www.reddit.com/r/starbound/comments/eowqjs/guns_not_functioning_frackin_universe_mod/
- https://steamcommunity.com/app/211820/discussions/4/2626095078191841404/

So maybe it is weapon stats...Anyway I hope we can give this an end.

sayterdarkwynd:
Im not interested in nonsense bugs from 5 years ago.

Do you have Weapon Stats installed? If so, that is your issue, not FU. It is not compatible with our mod.

linonetwo:
Sorry for wasting your time, it was "Weapon Retrofitting Station - Breaks weapons" in the list.

It says it is compatible in its page, but is not.

PR #3634, opened 2025-03-28
Update moogle.raceeffect

AmazonValkyrie:
Update to Moogle race effects from internal name WIP: The Viera of Ivalice - FU. Could this be added to FU's includes?


sayterdarkwynd:
It sure could, once you properly do it without syntax errors

PR #3638, opened 2025-04-05
Update moogle.raceeffect

AmazonValkyrie:
Update to Moogle race effects


AmazonValkyrie:
Update to Moogle race effects from internal name WIP: The Viera of Ivalice - FU. Could this be added to FU's includes?

PR #3639, opened 2025-04-05
Update moogle.species.patch

AmazonValkyrie:
Update to Moogle species patch reflecting updated race effects


AmazonValkyrie:
Update to Moogle species patch from internal name WIP: The Viera of Ivalice - FU. Could this be added to FU's includes?

PR #3640, opened 2025-04-05
Update viera.raceeffect

AmazonValkyrie:
Updated Viera race effects


AmazonValkyrie:
Update to Viera race effects from internal name WIP: The Viera of Ivalice - FU. Could this be added to FU's includes?

PR #3641, opened 2025-04-05
Update viera.species.patch

AmazonValkyrie:
Update to Viera species patch mirroring updated race effects


AmazonValkyrie:
Update to Viera species patch from internal name WIP: The Viera of Ivalice - FU. Could this be added to FU's includes?

PR #3687, opened 2026-04-19
Changed some things in research

Kukagres:
Floran furniture uses bamboo, not gold. That's why it makes more sense that it requires bamboo.
Anything that needs Densinium should be at least tier 6. The Biochemist table is tier 2, and since most slime items are made there, it makes more sense for it to be tier 2. Finally, Morphite is tier 4, and since it requires Morphite, that's why I listed it as tier 4.


sayterdarkwynd:
and where is a new player on a garden world meant to get bamboo?

Kukagres:
Advanced Crops -> Unlocks micro-bamboo -> get Bamboo Shots -> get bamboo

PR #3699, opened 2026-09-19
Implement displaying different text if an entry is researched vs. not researched

EtiTheSpirit:
This is kind of an out-of-nowhere change that was not previously discussed, sorry for the surprise.

This feature allows specifying a third string in a research strings element, changing the form to `['Title', 'Description', 'Optional Researched Description']`. As implied, it is displayed when the node has been researched. This change is transparent to existing research trees, modded or otherwise, and so there are no side effects to existing content.

Why? Conventional descriptions of research nodes are written as if the research is done, speaking about the results in the past tense. In my mod's custom research tree, I have a mystery questline that runs in parallel to the player's research. Obviously, talking about something that has yet to be done in the past tense is really not ideal for this kind of presentation. This change allows me to provide a user with a summary of their work once they've done it, without spoiling it before they do.

Hidden nodes do not currently address this specific use case, since they obscure the *children* of the unresearched node, where I need to obscure details from the *current* research node instead. That being said, I imagine this feature could be useful to make Metaphysics a bit more cryptic if you really wanted to, but aside from that I admittedly see little use case within current FU content.

Thanks for your time.


sayterdarkwynd:
need to fix the error in there.

EtiTheSpirit:
I did, I thought. Said PR description can't contain double quotes, and I fixed that.

Edit: I got a hunch, I'll turn it into a draft then reopen it.
Edit 2: There we go. Had to completely close and reopen.

